PR #1, opened 2019-06-21
Initial commit with nodes, nodes-express, java-microprofile and java-spring-boot2 stacks

neeraj-laad:
Initial commit with nodes, nodes-express, java-microprofile and java-spring-boot2 stacks
All stacks are at 0.2.0 version


README.md
```diff
@@ -1 +1,42 @@
...
\ No newline at end of file
+# Stacks
+
+This respository holds the default set of stacks avalible from the appsody cli. Stacks allow for rapid development whilst giving the stack developer the ability to control the overall applications that are created from it. A stack is make up of a stack and one or more templates.
+
+**Stack:** A stack is the foundation for your application. A stack can set default functionality and controls what can be overwritten by a template or application based on it.
+
+**Template:** A template is based upon a stack providing a template application that is ready to use. It has access to any dependencies supplied by that stack and is able to include extra dependencies plus new functionality to enhance the stack where allowed.
+
+## Getting started with Stacks
```
kylegc: This is overlap with some of the doc pages. We'll have to keep it consistent and in sync or maybe think about linking to the other docs instead.

PR #9, opened 2019-06-25
Enable code completion for Express

seabaylea:
This PR requires the users `app.js` to create their own Express.js app instance and export it using `module.exports.app = app;`.

This has two effects:
* It makes it much easier to bring an existing app - you just need to remove the `app.listen()` call and export the app instead
* It enables code completion for the use of Express APIs.


incubator/nodejs-express/image/project/server.js
```diff
@@ -18,13 +19,15 @@ function getEntryPoint() {
...
 }
 
-require(basePath + getEntryPoint())(app);
-
 const healthcheck = new health.HealthChecker();
 app.use('/live', health.LivenessEndpoint(healthcheck));
 app.use('/ready', health.ReadinessEndpoint(healthcheck));
 app.use('/health', health.HealthEndpoint(healthcheck));
 
+require(basePath + getEntryPoint()).app;
```
neeraj-laad: do we need the require twice?

PR #32, opened 2019-06-27
[java-spring-boot2]: final from adoptopenjdk; fix missing xmlstarlet

ebullient:


neeraj-laad:
Have built and pushed the new stack image.

PR #37, opened 2019-07-01
[java-spring-boot2]: Fix version update 

ebullient:
This supersedes PR #11

### DIFF
--- a/incubator/java-spring-boot2/README.md
+++ b/incubator/java-spring-boot2/README.md
@@ -6,7 +6,7 @@ The Spring Boot 2 stack uses a parent Maven project object model (POM) to manage
 
 The [Spring Developer Tools](https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-devtools.html#using-boot-devtools) module is also included to support automatic restarts and live reloading during development.
 
-*Note:* Maven is provided by the Appsody stack container, allowing you to build, test, and debug your Java application without installing Maven locally. We recommend installing Maven locally for the best IDE experience.
+**Note:** Maven is provided by the Appsody stack container, allowing you to build, test, and debug your Java application without installing Maven locally. We recommend installing Maven locally for the best IDE experience.
 
 ## Templates
 
--- a/incubator/java-spring-boot2/README.md
+++ b/incubator/java-spring-boot2/README.md
@@ -6,6 +6,8 @@ The Spring Boot 2 stack uses a parent Maven project object model (POM) to manage
 
 The [Spring Developer Tools](https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-devtools.html#using-boot-devtools) module is also included to support automatic restarts and live reloading during development.
 
+*Note:* Maven is provided by the Appsody stack container, allowing you to build, test, and debug your Java application without installing Maven locally. We recommend installing Maven locally for the best IDE experience.
+
 ## Templates
 
 Templates are used to create your local project and start your development.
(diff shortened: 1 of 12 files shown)


incubator/java-spring-boot2/README.md
```diff
@@ -6,6 +6,8 @@ The Spring Boot 2 stack uses a parent Maven project object model (POM) to manage
 
 The [Spring Developer Tools](https://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-devtools.html#using-boot-devtools) module is also included to support automatic restarts and live reloading during development.
 
+*Note:* Maven is provided by the Appsody stack container, allowing you to build, test, and debug your Java application without installing Maven locally. We recommend installing Maven locally for the best IDE experience.
```
neeraj-laad: Could we make the 'Note' heading bold? The other stacks have had the same issue and wanted to make this a bit more prominent

incubator/java-spring-boot2/image/Dockerfile-stack
```diff
@@ -11,30 +11,27 @@ RUN  apt-get -qq update \
...
 
-WORKDIR "/project"
-
-RUN mkdir -p /project/target && chmod 777 /project/target
+WORKDIR /project/user-app
 
 ENV APPSODY_USER_RUN_AS_LOCAL=true
 
-ENV APPSODY_MOUNTS="src:/project/src;pom.xml:/project/pom.xml;~/.m2/repository:/mvn/repository"
+ENV APPSODY_MOUNTS=".:/project/user-app;~/.m2/repository:/mvn/repository"
```
neeraj-laad: Interestingly the other stack( java-microprofile) went from mounting a single workspace dir to mounting src and pom individually as there were issues with creation of file / folders  that were in the directory that was bind mounted on the host. Were there specific reasons for changing the way the folders are mounted?

incubator/java-spring-boot2/image/Dockerfile-stack
```diff
@@ -11,30 +11,27 @@ RUN  apt-get -qq update \
...
 ENV APPSODY_DEPS=
 
-ENV APPSODY_WATCH_DIR="/project"
-ENV APPSODY_WATCH_IGNORE_DIR="/project/target"
+ENV APPSODY_WATCH_DIR="/project/user-app/src"
+ENV APPSODY_WATCH_IGNORE_DIR=""
 ENV APPSODY_WATCH_REGEX=""
```
neeraj-laad: Would you want to use any filters like only watch for changes to *.java files?

incubator/java-spring-boot2/README.md
ebullient: 👍

incubator/java-spring-boot2/image/Dockerfile-stack
ebullient: This is the most significant change, and is what allows the version plugin to run correctly. Remember that the Spring stack runs as user, so any files we create have proper ownership.  I’m not sure the microprofile stack does that yet.. that may have been a different way to fix the issue they were having.

incubator/java-spring-boot2/image/Dockerfile-stack
ebullient: no. Spring has other resource files that should also be watched.

incubator/java-spring-boot2/image/Dockerfile-stack
neeraj-laad: Yes, run as user is one of the fundamental differences between the 2 stacks.. and maybe it is good to have two different approaches so we can compare and contrast experiences from both before going full on on one.

Issue #30, opened 2019-06-27
Java-Microprofile: No starter app on the default route!


lroseblade:
Once the stack is installed and the user goes to https://locahost:9080 they receive a 404 error as there is no starter app defined and assigned to the end point.

neeraj-laad:
@Emily-Jiang Would like to know your thoughts on this issue.

barecode:
Proposing a change in a PR shortly...

barecode:
Proposed a small PR to make the default context root behave a little more user friendly. This does change the binding in the app away from /starter to /, but I think its an improvement in the initial user experience... especially when they are validating that they have appsody working properly.

neeraj-laad:
This has been resolved in version 0.2.2. We now have a single index.html that users can remove if needed.

PR #47, opened 2019-07-02
java-microprofile: make the default context root respond nicely

barecode:
java-microprofile: make the default context root (localhost:9080) respond with something better than a 404

This change adds a simple Hello World resource to the MicroProfile starter and bind the application to the default context root so that it responds on localhost:9080. When accessed, the user will see "Hello from Appsody!"

Fixes: https://github.com/appsody/stacks/issues/30

### DIFF
--- a/incubator/java-microprofile/templates/default/src/main/java/dev/appsody/starter/HelloResource.java
+++ b/incubator/java-microprofile/templates/default/src/main/java/dev/appsody/starter/HelloResource.java
@@ -1,0 +1,17 @@
+package dev.appsody.starter;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("")
+public class HelloResource {
+
+	@GET
+	@Produces(MediaType.TEXT_PLAIN)
+	public String getHelloWorld() {
+		return "Hello from Appsody!";
+	}
+
+}
--- a/index.yaml
+++ b/index.yaml
@@ -10,11 +10,11 @@ projects:
     - Microprofile
     - Maven
     - Open Liberty
-    name: java-microprofile
+    name: java-microprofile-mct
     maintainers:
     - neeraj.laad@gmail.com
     urls:
-    - https://github.com/appsody/stacks/releases/download/0.2.1/incubator.java-microprofile.templates.default.tar.gz
+    - https://github.com/barecode/stacks/releases/download/0.2.1/incubator.java-microprofile.templates.default.tar.gz
     version: 0.2.0
   java-spring-boot2:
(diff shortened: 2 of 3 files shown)


neeraj-laad:
@barecode  Welcome to Appsody! and thanks for your contribution.

We will review the PR shortly.

@Emily-Jiang would like to have your thoughts on this too.

incubator/java-microprofile/templates/default/src/main/java/dev/appsody/starter/HelloResource.java
```diff
@@ -0,0 +1,17 @@
+package dev.appsody.starter;
```
Emily-Jiang: Graham Charters and I discussed this when we first agreed on what to add. We decided not to add this one as this file will be deleted first. We just added something absolutely needed.

index.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 generated: 2019-06-24T21:00:00+00:00
 projects:
-  java-microprofile:
+  java-microprofile-mct:
```
Emily-Jiang: Why do we need to add -mct?

neeraj-laad:
@barecode Thanks for your contribution.

This looks like a good idea to keep all stacks consistent. Will get relevant people to review shortly.

incubator/java-microprofile/templates/default/src/main/java/dev/appsody/starter/HelloResource.java
Emily-Jiang: I had a chat with Graham again this morning. According to our survey, developers don't want a lot of code to be generated as they have to delete. This file will be the first one to be deleted. I suggest not to add in. This stack is for production use. We need to keep it concise and useful. Therefore, I suggest not to add this file.

index.yaml
neeraj-laad: The `mct` should be removed and version updated for the same stack.

neeraj-laad:
@barecode Are you able to update this PR by:

- remove `mct` references
- update version numbers in stack.yaml

@Emily-Jiang One one hand I agree that we do not want to clutter the developer's workspace, but on the other hand we should provide some feedback / success message when hitting the default endpoint.

We are looking for a way to return something meaningful on the default endpoint to make user feel that they have the application up and running. Developers should be able to remove that single file or comment a couple of lines to remove this. 

- Node and Nodejs-express stacks use a simple plan text response - "Welcome from Appsody"
- Spring stack uses an index.html with a welcome message, some links to Spring and a list of URLs.

Let me know what your thoughts are.

neeraj-laad:
This has been resolved in 0.2.2 version. Thanks for your contribution.

Please feel free to reopen this if you believe the issue has not been completely resolved.

Issue #31, opened 2019-06-27
[java-spring-boot2] No starter app on the default route! 


lroseblade:
A more consumable experience post install would be to have a 'Welcome to appsody!' page on the default route. This gives confidence the install has completed successfully.

ebullient:
We had the standard IBM Cloud landing page for apps.. do you want to resurrect / rebrand that?

lroseblade:
As long as the user who has just installed appsody gets an obvious visual confirmation that the install has been successful then that's great. While the 404 page is technically correct as there is no app assigned to the route, it can also give an impression the install perhaps didn't fully complete.

The node stacks complete with a "Hello World!" page, which I think is also fine (although 'Welcome to appsody stacks') would probaby be better. Whichever page is chosen it should be consistent throughout all the stacks.

seabaylea:
I think all we need is for `/` to respond with a simple "Hello from Appsody"/"Welcome from Appsody" so that a success is reported (rather than "Not Found" which might make people who are just evaluating the stack think that it's broken.

barecode:
Proposed a small PR to make the default context root behave a little more user friendly - following @seabaylea 's recommendation. I've also created PRs #47 and #49 to make the Hello world response "Hello from Appsody!" so its clear that the response is coming from an Appsody stack.

PR #48, opened 2019-07-02
[java-spring-boot2]: make the default context root respond nicely

barecode:
This change adds a RestController which responds on localhost:8080 with "Hello from Appsody!"

Fixes: https://github.com/appsody/stacks/issues/31

### DIFF
--- a/incubator/java-spring-boot2/templates/default/src/main/java/application/HelloController.java
+++ b/incubator/java-spring-boot2/templates/default/src/main/java/application/HelloController.java
@@ -1,0 +1,13 @@
+package application;
+
+import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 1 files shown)


neeraj-laad:
Thanks for your contribution. 

@ebullient, would like to get your thoughts on this.

ebullient:
Code is fine, but it is in the way of anyone following a guide.
I would prefer a simple index.html file. As soon as a controller is added, it would be replaced.

e.g. src/main/resources/public/index.html containing 
```
<html>
<head>
<title>Hello from Appsody</title>
</head>
<body>
Hello from Appsody
</body>
</html>
```

ebullient:
I will replace w/ a simple index.html file, as that is easier to replace/ignore than a controller

PR #49, opened 2019-07-02
[node-express]: Set the hello world message to be "Hello from Appsody!"

barecode:
This does not fix a particular issue, but makes the hello message for node-express consistent with the proposed changes to java-microprofile and java-spring-boot2


neeraj-laad:
Will wait for more changes (under review) to be merged and release them all together for this stack.

PR #55, opened 2019-07-04
[nodejs] Add appmetrics-dash as a dev dependency

seabaylea:
The Eclipse Codewind project uses the `appmetrics-dash` NPM module to provide development time application monitoring and performance testing data.

This updates the nodejs and nodejs-express stacks to add appmetrics-dash as a devDependency.

For the nodejs stack, this is done using `attach()`, which means it will only be added if the app provides some form of HTTP server. This is deliberate so that adding the dashboard doesn't prevent applications that are batch jobs etc from exiting.

### DIFF
--- a/incubator/nodejs/image/project/package.json
+++ b/incubator/nodejs/image/project/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nodejs",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "Node.js Stack",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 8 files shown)


seabaylea:
@neeraj-laad Updates made.

incubator/nodejs/image/project/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "nodejs",
+  "version": "0.2.0",
```
neeraj-laad: This should be 0.2.1 to keep it in line with the stack version

seabaylea: Done

seabaylea:
@neeraj-laad updates done

PR #56, opened 2019-07-06
[java-microprofile] Add javametrics as a dev only dependency

seabaylea:
This adds Javametrics-dash as a dev only dependency as part of the stack. This both provides developers with a build-in performance view, and enables the Eclipse Codewind IDE project to do application monitoring and performance testing the of the application during development.

The approach to providing the Javametrics dashboard during development only is as follows:
* Provide a pom-dev.xml file that installs the dependencies as installs them into the server which is only run in Dockerfile-stack, and not in the production Dockerfile.
* Provide a server-dev.xml that adds the necessary features that is only copied into the config dropins the Dockerfile-stack, and not in the production Dockerfile.

This pattern can be extended if there's further capabilities that we want to provide only during development.

### DIFF
--- a/incubator/java-microprofile/README.md
+++ b/incubator/java-microprofile/README.md
@@ -6,6 +6,8 @@ The Java MicroProfile stack uses a parent Maven project object model (POM) to ma
 
 This stack is based on OpenJDK with container-optimizations in OpenJ9 and `Open Liberty v19.0.0.8`. It provides live reloading during development by utilizing `loose application` capabilities.
 
+The stack also provides a in-built application monitoring dashboard based on [javametrics](https://github.com/runtimetools/javametrics). This dashboard is only included during development and is not included in the image built using `appsody build`.
+
 **Note:** Maven is provided by the Appsody stack container, allowing you to build, test, and debug your Java application without installing Maven locally. However, we recommend installing Maven locally for the best IDE experience.
 
 ## Templates
@@ -65,6 +67,7 @@ OpenAPI endpoints:
     - Metrics endpoint: http://localhost:9080/metrics (login as `admin` user with `adminpwd` password)
     - OpenAPI endpoint: http://localhost:9080/openapi
     - Swagger UI endpoint: http://localhost:9080/openapi/ui
+    - Javametrics Dashboard endpoint: http://localhost:9080/javametrics-dash/ (development-time only)
 
 ## License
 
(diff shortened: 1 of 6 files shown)


neeraj-laad:
@Emily-Jiang Would appreciate your feedback on this.

Emily-Jiang:
@seabaylea I have 2 comments
1. java-metrics feature for dev: I am not aware of how popular to have this low level java-metrics in dev. I think this info might just be useful to performance team. I don't think Jen is interested in this. If you think some developers might need it, I think it might fit in well to have this as a guide.
2. In pushing for dev-prod parity, I think it is not very useful for dev to have one tool while Prod having another tool to view different info. I think it is better to use MicroProfile metrics and have Prometheus and/or Grafana accessible to the Dev.

nastacio:
@seabaylea and @Emily-Jiang , this is also a matter of consistency across stacks when working with Codewind.

We ran into this during a workshop and writing guides, and pulled the PR into a forked repo just to showcase the functionality.

csantanapr:
I just hit this problem
Using Cloud Pak Apps stack for microprofile I got the impression that codewind is something to leverage in both java and nodejs stacks, using the App Monitor and Perf Dasboard launchers in codewind.

nodejs works fine but in java-microfile I hit the problem and opened an issue here 
https://github.com/eclipse/codewind-java-profiler/issues/8

ianpartridge:
I've rebased this PR so it no longer conflicts with `master`. It needs re-review by the stack owners.

@Emily-Jiang @gcharters @BarDweller @neeraj-laad please take a look

ianpartridge:
@ yeekangc too

gcharters:
I think this should be merged (or an updated version of it if the stack has changed since it was created) as without this capability the stack is not usable in Eclipse Codewind.  I consider this a tactical workaround and believe Codewind should be using the /metrics (e.g. MicroProfile metrics) endpoint from stacks that provide it.

Kamran64:
Just tested this and it works i.e I can get the Javametrics dashboard endpoint - However, I'll leave the approval to one of the reviewers mentioned above ^

Emily-Jiang:
This PR can be merged with the following agreement:

- This is a temp workaround. The ultimate solution should rely on MicroProfile Metrics. This PR does not demonstrate ecosystem nor the programming model.

- This PR should be removed once https://bigblue.aha.io/features/MC-156 is in place. 
I abstain from the voting.

ianpartridge:
@Emily-Jiang I edited your comment to remove the reference to an internal link. This is a public open source project.

yeekangc:
+1 to what Graham and Emily said.  This is a tactical solution and Codewind should use /metrics down the road.

ianpartridge:
@Kamran64 any update on the testing?

Kamran64:
@ianpartridge - @neeraj-laad and I took another look earlier. For some reason the command `'-javaagent:resources/javametrics-agent.jar'` fails only on `appsody test` but not on `run` and `debug`. We think this could probably be removed from `APPSODY_TEST` env in the Dockerfile-stack though.

PR #82, opened 2019-07-15
Add swift to index

neeraj-laad:
Signed-off-by Neeraj Laad neeraj.laad@gmail.com

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -57,7 +57,7 @@ projects:
     version: 0.2.1
   swift:
   - created: 2019-07-15T12:00:00+00:00
-    description: Swift application
+    description: Swift runtime
     icon: https://
     keywords:
     - Swift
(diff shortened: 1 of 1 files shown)


index.yaml
```diff
@@ -54,4 +54,16 @@ projects:
     - neeraj.laad@gmail.com
     urls:
     - https://github.com/appsody/stacks/releases/download/nodejs-express-v0.2.1/incubator.nodejs-express.templates.simple.tar.gz
-    version: 0.2.1
\ No newline at end of file
+    version: 0.2.1
+  swift:
+  - created: 2019-07-15T12:00:00+00:00
+    description: Swift application
```
mhamwala: I think we chose to go for Swift runtime in the website PR, similar to the Node.js stack

neeraj-laad: good catch.. fixed.

PR #86, opened 2019-07-15
Use mvn package instead of compile on change

kylegc:
- [ ] commit message follows [commit guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md#commit-message-guidelines).

## Related Issues

Fixes #
Related to https://github.com/appsody/controller/issues/7

### Description

I propose using `mvn package` rather than `mvn compile` in the java-microprofile stack ON_CHANGE command. 
The reason is that `package` allows liberty to pick up config changes to server.xml such as creating new endpoints or features. In an obscure case (which we hit testing https://github.com/appsody/controller/issues/7 ) if the liberty server is started with a compile error in one of the endpoints, fixing the compile error and running `mvn compile` does not enable the endpoint but `mvn package` does.

See also https://openliberty.io/guides/getting-started.html#updating-the-server-configuration-without-restarting-the-server

The downside of this is a longer ON_CHANGE command. On my local machine it takes roughly 3 secs to run package vs 1 sec to run compile.


neeraj-laad:
@Emily-Jiang Would like to know what you think of this change? The idea is to allow users to change liberty config and let appsody pickup those changes too.

PR #101, opened 2019-07-22
kotlin-spring-boot2: adding a Kotlin with Spring Boot 2 stack

pushkarnk:
This is a refinement of #76 

### Checklist:

- [x] Read the [Code of Conduct](https://github.com/appsody/website/blob/master/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md).

- [x] Followed the [commit message guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md#commit-message-guidelines).

- [x] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stack-structure.md).

### Contributing a new stack:

- Describe how application dependencies are managed:
This PR relates to Kotlin support with Spring Boot 2. Just like the `java-spring-boot2` stack the dependency management is done using Maven's parent object model (POM).

- Explain how Appsody file watcher is utilised:
The `APPSODY_WATCH_REGEX` is, for now, set only for `.kt` files. This triggers a recompile whenever a Kotlin source file is modified.

- Describe other Appsody environment variables defined by the stack image:
This defines only those environment vars that are defined by `java-spring-boot2`.

- Describe any limitations and known issues:
This stack has a structure that is identical to `java-spring-boot2`. If `appsody` provided a way to inherit code from other stacks, we could be reducing on the LOC here.

PR #107, opened 2019-07-23
Add Kotlin Template to Spring Boot Stack

BarDweller:
### Checklist:

Adds Kotlin compiler to the base Spring Boot 2 stack. 
Adds Kotlin Template to the Spring Boot 2 stack. (Pending resolution of Issue https://github.com/appsody/appsody/issues/38 )

Note there are some gymnastics here to enable Kotlin version to be defined outside the Dockerfiles, ideally the version should also be used to generate? the parent pom, but for now the cleanest route is to just have the version defined in two places. This places an added burden on the Stack author to ensure the version remains in sync in both places, but since a human is involved, maintains a lower risk of accidentally creating multiple parent poms with the same maven coordinates (group/artifact/version) but with different content (this would be bad).

### Related Issues:
Related to PR#101, but adds Kotlin to the base boot stack, with the intent of making it available as a template, rather than as a parallel stack.


pushkarnk:
@BarDweller Though its not a "must-do", the tests could be in Kotlin too. You may refer to the changes of #101

Issue #127, opened 2019-07-31
Init process different between Java and Spring stacks


makandre:
**Describe the bug**

I’ve observed that the `init` for Java and Spring stacks run differently.

For Java, `init` runs `mvn` locally and setup the m2 cache such that I can do `mvn` commands in the project right after.  

For Spring, `init` only setup the m2 cache after an `appsody run` (it appears setup happens inside the app container where `~/.m2/repository:/mvn/repository` mount exists).  As a result, running `mvn` commands in spring project after `init` gives “Non-resolvable parent POM” errors.

**Additional context**

I'm coming at this as a developer for the appsody extension for Codewind.  It is preferable for us that the initialization behaviour is consistent between the stacks so we can use a single method to configure the IDEs correctly (VSCode and Eclipse) when creating an appsody project.

Also, the Java way of initialization essentially requires for `mvn` to be present (aside: isn't appsody supposed to shield the developer from requiring specific tools?)   This is problematic for us too because we run appsody in Codewind's container (DinD) and we likely cannot install maven.

Spring way of initialization works better for us, but happening during `run` is not ideal because when the project is imported into Eclipse workspace after it's created and it's supposed to be "initialized" but it is actually not, so the project will not be configured correctly.

**To Reproduce**

**Java:**
1. `appsody init java-microprofile`
2. `mvn validate`
3. Succeeds

**Spring:**
1. `appsody init java-spring-boot2`
2. `mvn validate`
3. Get "Non-resolvable parent POM" errors
(need to run `appsody run` in order for m2 cache to be setup properly)

**Expected behavior**

The behaviour between Java and Spring initialization should be consistent

**Actual behaviour**

The behaviour between Java and Spring differs as described above

**Environment Details (please complete the following information):**

**If applicable please specify:**
 - CLI version:
 - Stack you are using:

**Screenshots**

neeraj-laad:
@makandre You are right there are 2 slightly different approaches in those 2 stacks. The main reason being the init hooks were not present when the spring stack was initially created. We can look at implementing that, but there are always pros and cons to each approach and we would like the stack providers to be able to choose what they believe works best for them and their developers.

Also, the maven dependency (incase of microprofile) is only needed for IDEs to be happy. When users execute `appsody run` the appsody container provides all that is needed to run the application.

@ebullient @Emily-Jiang Would like your thoughts on this.

makandre:
[...]

PR #162, opened 2019-08-09
java-microprofile: Maven cache share

neeraj-laad:
### Checklist:

This change enables maven cache sharing between Appsody container and host running appsody and allows the parent to be installed during appsody run as well. 

There is an assumption that cache is in ~/.m2/repository, and there is no way to change it today.

Incase of DinD (e.g CodeWind), an environment variable called APPSODY_MOUNT_HOME can be set before invoking `appsody run/test/debug`. and the value of that variable will be use instead of using `~`. for sharing cache.

### Related Issues:
fixes: #127


incubator/java-microprofile/templates/default/.appsody-config.yaml
```diff
@@ -1 +1 @@
-stack: appsody/java-microprofile:0.2
+stack: appsody/java-microprofile:0.2.5
```
mhamwala: Let's keep this at **0.2** to pick up the latest

neeraj-laad: good catch.. fixed

Issue #161, opened 2019-08-09
Generate index.json for CodeWind to consume appsody projects


neeraj-laad:
Please see a sample of index in https://github.com/appsody/stacks/pull/131

PR #163, opened 2019-08-09
stacks: generated index.json for CodeWind

skoh7645:
### Checklist:

### Description
Wrote a python script to generate an index.json file from the index.yaml for CodeWind to consume appsody projects. The python script is called in the build.sh script.

### Related Issues:
Fixes: #161


ci/yaml_to_json.py
```diff
@@ -0,0 +1,50 @@
...
+        try:
+            doc = yaml.safe_load(s)
+            i = 0
+
+            if (doc['stacks'] != None):
+                for item in doc['stacks']:
+                    res = (OrderedDict([
+                        ("displayName", item['name']),
+                        ("description", item['description']),
+                        ("language", item['id']),
```
neeraj-laad: language should be set to blank for now: "" as there is no concept of language in our stack.yamls

ci/yaml_to_json.py
```diff
@@ -0,0 +1,50 @@
...
+                        ("projectType", "appsodyExtension"),
+                        ("projectStyle", "Appsody"),
+                        ("location", item['templates'][0]['url']),
```
neeraj-laad: I believe you should pull in url for the default-template here

ci/yaml_to_json.py
makandre: Unfortunately Codewind does require a language.  Could we hardcode this for now to one of `java`, `nodejs`, `swift` accordingly?

ci/yaml_to_json.py
skoh7645: Yes, I have fixed this in the latest commit

ci/yaml_to_json.py
neeraj-laad: @makandre it is not possible for us to set these on the fly, unless we maintain a map of stacks and languages somewhere (quarkus -> java).

As per @tobespc, "" would suffice. Please, can you confirm?

makandre: Toby's away but maybe @deboer-tim can comment?

makandre: Although we can create a project w/o the language set, we will fail to display it in several places within the IDE as you can see in the screenshot:

image

It also breaks debugging.  Codewind uses the `language` to determine the style of debugging to use.

deboer-tim: Sorry, I don't know how the language is used and whether it gets through the template API to the client. I'm assuming at a minimum it would mean you can't filter the list of projects by language (which isn't necessarily a huge loss) but I don't if it would cause other problems. Might be worth trying (at least to see if it is ok for now), or pinging Howard.

makandre:
Hi, I see this got merged.  Where is the `index.json` file located?

I also see the language field is going to be blank.  As I mentioned in https://github.com/appsody/stacks/pull/163#discussion_r313541209 this is going to break the debug scenario for Codewind.

PR #190, opened 2019-08-16
[Swift] Add apt-get zlib1g-dev for use by Vapor apps

seabaylea:
### Checklist:

- [X] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stack-structure.md).

- [X] Updated the stack version in `stack.yaml`

This adds `apt-get install -y zlib1g-dev` to `Dockerfile-stack` for development and to the production `Dockerfile` for use by Vapor based apps.

### DIFF
--- a/incubator/swift/image/Dockerfile-stack
+++ b/incubator/swift/image/Dockerfile-stack
@@ -1,6 +1,10 @@
 FROM swift:5.0
 
-RUN apt-get update && apt-get install -y libcurl4-openssl-dev libssl-dev
+RUN apt-get update \
+ && apt-get dist-upgrade -y \
+ && apt-get install -y zlib1g-dev libcurl4-openssl-dev libssl-dev \
+ && apt-get clean \
+ && echo 'Finished installing dependencies'
 
 ENV APPSODY_MOUNTS=/:/project/user-app
 ENV APPSODY_DEPS=/project/user-app/.build
--- a/incubator/swift/image/Dockerfile-stack
+++ b/incubator/swift/image/Dockerfile-stack
@@ -1,7 +1,6 @@
 FROM swift:5.0
 
 RUN apt-get update \
- && apt-get dist-upgrade -y \
  && apt-get install -y zlib1g-dev libcurl4-openssl-dev libssl-dev \
  && apt-get clean \
  && echo 'Finished installing dependencies'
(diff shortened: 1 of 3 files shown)


incubator/swift/image/project/Dockerfile
```diff
@@ -19,7 +20,8 @@ FROM swift:5.0-slim
 
 # Install OS updates
 RUN apt-get update \
- && apt-get install -y libcurl4-openssl-dev libssl-dev \
+ && apt-get dist-upgrade -y \
+ && apt-get install -y zlib1g-dev libcurl4-openssl-dev libssl-dev \
```
ianpartridge: I think we just need `zlib1g` not the dev package as this is the runtime slim section?  Should only need the dev headers for the build.

ianpartridge: Also, `libcurl4` and `libssl1.1`, maybe (haven't tested it).

ianpartridge: Also, while you're in here can we change `swift build -c release > output.txt` to `swift build -c release | tee output.txt` so the output goes to both stdout and the file. Otherwise the user can't see their app being built.

ianpartridge: Maybe we can even get rid of the `output.txt` and do something like `swift build -c release | tee awk 'END {print $NF}' >> run.sh` - dunno.

incubator/swift/image/Dockerfile-stack
```diff
@@ -1,6 +1,10 @@
 FROM swift:5.0
 
-RUN apt-get update && apt-get install -y libcurl4-openssl-dev libssl-dev
+RUN apt-get update \
+ && apt-get dist-upgrade -y \
```
ianpartridge: What's the reason for `dist-upgrade`? Docker say not to do this: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

seabaylea: Removed

incubator/swift/image/project/Dockerfile
seabaylea: I've added `tee`, but left it as two separate tasks.

PR #300, opened 2019-09-03
java-spring-boot2: Use the Docker stack image to build templates

ebullient:
Propagate Stack id, major, minor, patch versions into build
artifacts in the project (Dockerfile), and in the project templates (pom.xml)

Signed-off-by: Erin Schnabel <schnabel@us.ibm.com>

### Checklist:

- [ ] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stack-structure.md).

-- > This proposes an _optional_ change to the stack structure, allowing templates to be built by, and stored in, the stack itself.

- [X] Updated the stack version in `stack.yaml`

The Spring stack has been adjusted to move the templates inside the stack image (`.../image/templates`) to allow the templates to be rebuilt when the stack is rebuilt. Specifically: "building" the templates includes propagation of version and other stack attributes into the template text. The goal is to reduce the number of changes that need to be made to ensure consistent stack/template versioning and uniqueness. 

Attributes of the stack (like the specified git organization and repository name) are used in the Spring stack for things like the maven group and artifact ids for the parent pom, and the group id for the template. The knock on: if I fork the repository, the maven pom that is generated is unique to  my fork, and doesn't overlay other existing maven pom files.

The version specified in stack.yaml is used for the version of the parent pom, and generates the version ranges used in templates and in .appsody-conf.yaml files (the stack takes over creating these files, and ensures the correct version range is used)

### Contributing a new stack:


incubator/java-spring-boot2/image/project/Dockerfile
```diff
@@ -1,21 +1,12 @@
...
-LABEL maintainer="IBM Java Engineering at IBM Cloud"
-
-# Ensure up to date / patched OS
-RUN  apt-get -qq update \
-  && apt-get -qq install -y curl wget xmlstarlet unzip \
-  && DEBIAN_FRONTEND=noninteractive apt-get -qq upgrade -y \
-  && apt-get -qq clean \
-  && rm -rf /tmp/* /var/lib/apt/lists/* \
-  && mkdir -p /mvn/repository
+FROM {{APPSODY_STACK}} as compile
```
henrynash: So I assume this is attempting to replace the installation of the latest OS updates with simply using those from the stack image itself? While this seems a sensible thing, we should probably make sure we are all OK with as a general policy (to date, I think we have always stated that the two builds are independent). As an aside, I assume this change is not actually required for the template change...might it be better to do this in two separate PRs?

incubator/java-spring-boot2/image/Dockerfile-stack
```diff
@@ -10,17 +17,49 @@ RUN  apt-get -qq update \
...
+RUN if [ ${MAJOR_VERSION} -eq 0 ]; then \
+      export CONFIG_VERSION=${MAJOR_VERSION}.${MINOR_VERSION}; \
+    else \
+      export CONFIG_VERSION=${MAJOR_VERSION}; \
+    fi \
+ && for x in /templates/*; do \
+      cd $x; \
+      name=$(basename $x); \
+      echo "stack: ${IMAGE_REGISTRY_ORG}/${STACK_ID}:${CONFIG_VERSION}" > .appsody-config.yaml; \
+      tar -czf /templates/${name}.tar.gz --exclude-from=.gitignore .; \
```
henrynash: My personal preference would be for us to leave the full source of the templates in the stack image, leaving open the technique we use to install them to the init stage,

incubator/java-spring-boot2/image/project/Dockerfile
ebullient: This change cascades from the template change, as it is using variable substitution. The APPSODY_STACK is a variable in the Dockerfile that is cascaded down. I could pull it out, but I'd rather not. I think the behavior change all goes together.

As for updates: this is all about how something is _built_ .. not how it is run. The above changes the first stage (as compile), not the second stage (the run image). This means that the same stack will be used when building locally, and when building in the pipeline, which seems much more predictable and consistent to me.

incubator/java-spring-boot2/image/Dockerfile-stack
ebullient: I could easily leave the source for analysis.. but the tarball is needed by IDE and other tools. Extracting a directory of files is more work than extracting a tgz, it also has a greater impact on the release process.

incubator/java-spring-boot2/image/project/Dockerfile
henrynash: I tend to agree - keeping the stack the same is also good for ensuring you don't inadvertently introduce changes as you move from run->build->deploy

ebullient:
Closing -- will rebase and split into more digestible pieces (good feedback @henrynash)

PR #311, opened 2019-09-04
Allow the sample stack to be run by appsody

### LINKED ISSUE #310: Make the sample-stack executable by Appsody
The sample stack has already been improved thanks to the work of @Kamran64 . What would make it even more useful is to allow it to be actually executable by Appsody (i.e. you can run/build/deploy it). This would allow a new stack developer to copy the sample stack, run it to provide that everything is set up OK, and then start modifying it.

In terms of the application could be just something like "echo" :-)

henrynash:
Make the sample stack even more useful by allowing appsody to actually run/build/deploy it. This means a stack developer can ensure their dev environment is set up OK, that they can run the sample - ahead of starting to modify it.

Fixes #310


henrynash:
This PR will be replaced by one that provides an additional sample stack (that is executable), rather than modifying the existing one.

henrynash:
This PR is now superseded by #361

PR #352, opened 2019-09-16
python-flask: unset watch ignore directory

### LINKED ISSUE #350: Remove redundant ignore line in python-flask stack
The current python-flask stack has an APPSODY_WATCH_IGNORE_DIR defined for a directory that is not included in the stack. We should remove this.

henrynash:
There are no directories within the user application directory, so no need to define any watch ignore directories.

Fixes #350


henrynash:
yes, and we should (I wasn't sure if this variable was mandatory, but now realise that it is not). let me re-push the PR>

PR #362, opened 2019-09-18
Adding a resource to MP template

arthurdm:
The current Java MicroProfile stack starter has an incomplete JAX-RS skeleton.  The URLs that are available are only the Liberty-provided endpoints, such as `/openapi/ui`, `/health` and `/metrics`.

The actual application, by default at `localhost:9080/starter` gives the following error:

```
Error 500: javax.servlet.ServletException: At least one provider or resource class should be specified for application class "dev.appsody.starter.StarterApplication
```

The issue is that no resources are actually provided.  This PR fixes this issue by providing a simple one, which can be easily customized.  With this PR the application endpoint `localhost:9080/starter` now outputs:  `Microservice v1`

### DIFF
--- a/incubator/java-microprofile/templates/default/src/main/java/dev/appsody/starter/StarterResource.java
+++ b/incubator/java-microprofile/templates/default/src/main/java/dev/appsody/starter/StarterResource.java
@@ -1,0 +1,18 @@
+package dev.appsody.starter;
+
+import static javax.ws.rs.core.MediaType.TEXT_HTML;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+
+@Path("/")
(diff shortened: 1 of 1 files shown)


arthurdm:
CC @Emily-Jiang

neeraj-laad:
@arthurdm We specifically discussed this with @Emily-Jiang and @gcharters and the thought was that we did not want to provide users with code that they have to delete. Hence we did not add this.

But, if the stack maintainers are happy to add this then we are good to merge.

arthurdm:
thanks @neeraj-laad - I will defer to @Emily-Jiang / @gcharters then.   I do think though, that would be nice to have a functional JAXRS app out-the-box.

arthurdm:
You're correct @neeraj-laad - I will close this.

Issue #354, opened 2019-09-16
ci build scripts for a specific repo/stack seems to process all repos


henrynash:
When trying to build a specific repo/stack, the build scrips seem to do a lot more...e.g.

$ ./ci/build.sh incubator/python-flask
 == Running pre_env.d scripts
 == Done pre_env.d scripts
~/go/src/github.com/appsody/stacks/ci/build/prefetch ~/go/src/github.com/appsody/stacks
== https://github.com/appsody/stacks/releases/latest/download/experimental-index.yaml
   https://github.com/appsody/stacks/releases/download/java-spring-boot2-liberty-v0.1.9/experimental.java-spring-boot2-liberty.v0.1.9.templates.default.tar.gz
   https://github.com/appsody/stacks/releases/download/nodejs-functions-v0.1.3/experimental.nodejs-functions.templates.simple.tar.gz
   https://github.com/appsody/stacks/releases/download/quarkus-v0.1.5/experimental.quarkus.v0.1.5.templates.default.tar.gz
   https://github.com/appsody/stacks/releases/download/vertx-v0.1.2/experimental.vertx.v0.1.2.templates.default.tar.gz
== https://github.com/appsody/stacks/releases/latest/download/incubator-index.yaml
   https://github.com/appsody/stacks/releases/download/java-microprofile-v0.2.14/incubator.java-microprofile.v0.2.14.templates.default.tar.gz
   https://github.com/appsody/stacks/releases/download/java-spring-boot2-v0.3.13/incubator.java-spring-boot2.v0.3.13.templates.default.tar.gz
   https://github.com/appsody/stacks/releases/download/java-spring-boot2-v0.3.13/incubator.java-spring-boot2.v0.3.13.templates.kotlin.tar.gz
   https://github.com/appsody/stacks/releases/download/nodejs-express-v0.2.6/incubator.nodejs-express.v0.2.6.templates.simple.tar.gz
   https://github.com/appsody/stacks/releases/download/nodejs-express-v0.2.6/incubator.nodejs-express.v0.2.6.templates.skaffold.tar.gz
   https://github.com/appsody/stacks/releases/download/nodejs-loopback-v0.1.4/incubator.nodejs-loopback.templates.scaffold.tar.gz
   https://github.com/appsody/stacks/releases/download/nodejs-v0.2.5/incubator.nodejs.templates.simple.tar.gz
[...]

Issue #355, opened 2019-09-16
ci build script get confused if standard index is linked to more than one repo name


henrynash:
If you happen to already have a local repo pointing at one the standard local indexes, e.g.

```
local-test	file:///Users/henrynash/go/src/github.com/appsody/stacks/ci/assets/experimental-index-local.yaml
```
then the build scrips will fail to rebuild this index (since they appear to assume you want the repo to be called experimental-index-local and error out when they see that that index is already assigned to another local repo name) and the subsequent testing will fail since the index doesn't exist.

ebullient:
We went around a bit on this. As of now, I would let the scripts add/remove the local test repos for you (as they can now do). It is inferring the repo name for local indexes based on the directory structure + local to avoid conflicts. ..  For ci/test and other scripts, using a determinable/predictable name is preferred.

henrynash:
...and that's a fine approach...although what confused me were the docs on building stacks which seems to advise you manually indexes to private local repo names (see: https://appsody.dev/docs/stacks/build-and-test). Doesn't this kind of guarantee we push people into this problem?

neeraj-laad:
@Kamran64 can you please review this doc and update to match the current build / ci process.

henrynash:
As an aside, this is a permanent error (i.e. it doesn't go away if you run the ci build again - it stays broken). The only way to make the build work is to remove your other local repo name with `appsody repo remove.`

If there are good reasons why two repo names can't point at the same index, then we should document this restriction. However, I'm not sure why we should make such a restriction...and if we removed that restriction, then this who problem would go away.

neeraj-laad:
@groeges

neeraj-laad:
This should be resolved now as the CI scripts are no longer adding/removing repos. @groeges can you confirm this is accurate and close the issue?

groeges:
The new CI scripts do not use the test.sh anymore as they perform the stack validation (testing) using the `appsody stack validate` command. This uses an entry added to the dev.local-index.yaml under the ./appsody/stacks/dev.local directory which is added to the appsody repo list as dev.local
```
*incubator  	https://github.com/appsody/stacks/releases/latest/download/incubator-index.yaml        
dev.local   	file:///Users/myuserid/.appsody/stacks/dev.local/dev.local-index.yaml             
experimental	https://github.com/appsody/stacks/releases/latest/download/experimental-index.yaml     
```
[...]

PR #367, opened 2019-09-20
ci: only generate indexes for stacks being built

Kamran64:
### Checklist:

- Makes prefetch optional via an environment variable (default is off) - On for travis builds
- Generate indexes for only built stacks i.e. no need to generate experimental index when only incubator stacks are being built
- Only add repos (for ```appsody repo add```) for stacks being built.

### Related Issues:
Fixes: #354 
Related to: #355


ebullient:
There should be no TRAVIS specific logic in package.sh. That needs to be corrected.

ci/test.sh
```diff
@@ -113,7 +113,14 @@ test_template() {
     popd
 }
 
-for x in $assets_dir/*.yaml
+if [ "$TRAVIS" == "true" ]
```
ebullient: This is not a TRAVIS-only path.

ebullient: Even the travis build can (and should) use the -local index, as it has the path to the _built_ tarballs, rather than to those that haven't been published yet.

ci/package.sh
```diff
@@ -145,25 +181,31 @@ do
...
                             fi
 
-                            echo "      - id: $template_id" >> $index_file
-                            echo "        url: $RELEASE_URL/$stack_id-v$stack_version/$versioned_archive" >> $index_file
                         elif [ -f $prefetch_dir/$versioned_archive ]
                         then
-                            # Add references to existing template archive.
-                            echo "      - id: $template_id" >> $index_src
-                            echo "        url: {{EXTERNAL_URL}}/$versioned_archive" >> $index_src
+                            if [ "$TRAVIS" == "true" ]
```
ebullient: Do not use a TRAVIS env var in package.sh. This is valid in other environments besides just TRAVIS. There is a different way to indicate whether or not certain indexes should be built. In all cases, if the repo is built, the local template should be updated. All we need here is a flag indicating whether or not we need a full release index to be generated as well, and that can be guarded by the prefetch variable (as you need the prefetch contents to generate the full index, including things that were not previously built)

ci/test.sh
```diff
@@ -113,7 +113,14 @@ test_template() {
     popd
 }
 
-for x in $assets_dir/*.yaml
+if [ "$GENERATE_ALL" == "true" ]
+then
+    indexPath=$assets_dir/*.yaml
+else
+    indexPath=$assets_dir/*-local.yaml
```
ebullient: I think test should always go against *-local.yaml. I think I got this wrong in the first pass.

*-local.yaml files will only exist for content that was generated.. so this should self-limit (and could correct some other problems with automatic addition of repositories)

Kamran64: Just for clarification, are you saying ```indexPath``` should always be *-local.yaml regardless of the value of ```GENERATE_ALL```?

neeraj-laad:
Closing this PR as this might is solved in a different way by `appsody stack package` for stack developers.  We can revisit this when the ci has been updated to use those.

PR #406, opened 2019-10-04
Fix Build Script

EnriqueL8:
### Checklist:

- [X] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stack-structure.md).

Fixing the build script to correctly build the stacks. In order to change export variable in a bash script, they need to created with source as such: https://stackoverflow.com/questions/13796809/changing-the-value-of-a-export-variable-from-a-bash-script


Kamran64:
Thanks @EnriqueL8 - this makes sense to me but I think there's a reason why those extension hooks are only running if ```STACKS_LIST``` is empty. Maybe we should have a double check from @neeraj-laad

EnriqueL8:
Actually @Kamran64, we could just put the all the logic related to `STACK_LIST` inside the `if` block which exports `STACK_LIST` itself. This would mean that we can still use the `-z` check.

PR #429, opened 2019-10-16
rust: support for debugging

skoh7645:
### Checklist:

Added low level debugging functionality for the rust stack


experimental/rust/README.md
```diff
@@ -67,6 +65,16 @@ You can enable an existing project as follows:
 
 3. You should see your application run, with logs written to the console.
 
+## Debugging
+
+You can run also your application in debug mode using the following command:
+
+```bash
+    appsody debug --docker-options "--cap-add=SYS_PTRACE --security-opt seccomp=unconfined"
```
ianpartridge: Can we push these Docker options down into the stack definition? e.g. an `APPSODY_DEBUG_DOCKER_OPTIONS` env on the stack image.

skoh7645: I thought the use case wouldn't be strong enough as this is only required in the debug command of the rust stack as of now. However, if we consider the other possible stacks with this problem (e.g. swift, rocket...) this could be a valid future requirement. Thanks for clarifying that this would be considered a "future" requirement.

PR #485, opened 2019-11-14
java-microprofile : Cache maven dependencies during appsody build

### LINKED ISSUE #371: Use maven cache for docker build
**Is your feature request related to a problem? Please describe.**

Optimize building java applications in appsody by packing dependencies in the multi-stage build image. See https://github.com/halkyonio/operator/issues/149: 

> Observation: When a s2i or maven build will take place on the k8s platform, then the Maven Deps or GAVs needed to compile/package a java runtime as a uber jar will be downloaded for every build. This way of working is not really green/environmentally responsible and could be resolved by supporting a maven local cache

For our (appsody) purposes, this is specifically related to the artifacts used by the first stage of the multi-stage Dockerfile inside the stack image.

awisniew90:
### Checklist:

Break up the Dockerfile used during `appsody build` to allow caching of Maven dependencies. This will prevent redownloading all dependencies when source code in the user app is changed. 

### Related Issues:
Fixes #371


scottkurz:
The way I read #371 suggests using some caching mechanism of something like Travis.  

This looks, rather, like a smarter layering, reflecting the fact that every issuing of `appsody build` for a given stack will result in the same dependencies _coming from the parent POM__.

If you agree, maybe remove the #371 reference and the word "cache"?

neeraj-laad:
I have to agree I thought this will cache the open liberty zip files in the stack development image (Dockerfile-stack). It would be great if we can have the zip in the image instead of it downloading when the user runs `appsody run/test/debug`.

awisniew90:
@neeraj-laad - we can layer the Dockerfile-stack file as well to get a similar behavior when `appsody run` is executed.

awisniew90:
Looking at this more... in the dev image (Dockerfile-stack) we are mounting the .m2 directory into the container when `appsody run` is called, so we are only downloading the liberty zips once in the local workspace. Subsequent runs pull the zips from the local maven repo. I may be missing something in the run/test/debug lifecycle, but we are already caching the dependencies in this case.

scottkurz:
Reading over the comment from @neeraj-laad 

> I have to agree I thought this will cache the open liberty zip files in the stack development image (Dockerfile-stack). It would be great if we can have the zip in the image instead of it downloading when the user runs `appsody run/test/debug`.

Like @awisniew90 said, these are already cached in the local .m2 repo, though it could still be nice to improve the experience on the very first `appsody run` for this stack.   

I'm not sure I'd stop at having the zip in the image...  Should we just go ahead and do the `mvn liberty:install-server` so that it gets unzipped into the `target/liberty/wlp/`   ?    

And we could always cut this PR at some point and do more later.. but see what you think.

awisniew90:
Good point, we could expand the zip in a docker layer to improve time on later runs.

awisniew90:
I will open another issue to focus on enhancements to the `appsody run/test/debug` (dev path) and keep this PR strictly enhancements to the `appsody build` path.

PR #532, opened 2019-12-02
java-spring-boot2: publish application in layers

ebullient:
### Checklist:

- [x] Read the [Code of Conduct](https://github.com/appsody/website/blob/master/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md).

- [x] Followed the [commit message guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md#commit-message-guidelines).

- [x] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stack-structure.md).

### Modifying an existing stack:

- [x] Updated the stack version in `stack.yaml`

The stack's build instructions were modified to extract dependencies at build time. The final docker image is then constructed using those extracted dependencies.

The root vs. java user has been cleaned up in both the stack and final docker images.

### Related Issues:

Fixes #527

PR #543, opened 2019-12-04
WIP: java-spring-boot2: Spring boot layering.. 

BarDweller:
### Checklist:

Building from @ebullient 's #532 PR, this one autosets the final classname using the spring boot fat jar manifest. Name is extracted in package script step, and Dockerfile assembles it into a start.sh for the final container. It would be easier to do the entire script from the package step, but I wanted to try to keep the exec step visible in the dockerfile

Additionally the RUN step setting the layer chown is removed (because I think it would undo the benefit of having copied the app as layered). Instead, uid/gid is set in prior stage, and during COPY step. 

Also added some stuff to try to suppress the actuator e2e trace.

### DIFF
--- a/incubator/java-spring-boot2/image/project/Dockerfile
+++ b/incubator/java-spring-boot2/image/project/Dockerfile
@@ -12,6 +12,12 @@ WORKDIR /project/user-app
 RUN /project/util/check_version build \
  && /project/java-spring-boot2-build.sh package
 
+#Build start.sh
+RUN echo '#!/bin/sh' > /project/user-app/target/start.sh \
+ && cat /project/user-app/target/mainclass | sed -e 's#\(.*\)#sh -c "exec java $JVM_ARGS -cp /app:/app/lib/* -Djava.security.egd=file:/dev/./urandom \1"#' >> /project/user-app/target/start.sh \
+ && chmod +x /project/user-app/target/start.sh \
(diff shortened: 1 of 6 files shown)


incubator/java-spring-boot2/image/project/Dockerfile
```diff
@@ -3,17 +3,26 @@ FROM {{.stack.image.namespace}}/{{.stack.id}}:{{.stack.semver.major}}.{{.stack.s
 #setup project folder for java build step
 COPY . /project
 
+USER root
+RUN chown -R java_user:java_group /project
```
Kamran64: If we're using the `chown` flag, please could you add a `docker-version` constraint on the `requirements` section in the `stack.yaml`. Should be a value of ">= 17.09.0"

BarDweller: Aye, we were discussing this in the appsody slack (s'why this is just a WIP). 
We were wondering if it's better to find a different way to handle the chown, to retain compatibility with earlier Docker versions.
I need to investigate the impact to layering for 
 - using chown flag on each copy step
 - using individual run chown steps after each copy step
 - using a compound run chown step after all copy steps
The concern is the run chown steps will affect every file from the copy, which could negate the effect of layering the image (smaller uploads when only code is changed, not deps/libs etc).

ebullient:
Closing this in favor of #559

PR #567, opened 2019-12-13
Added SCC option for java stack for a better startup

bharathappali:
**STACK : java-spring-boot2**

As class data sharing improves the JVM startup time, This PR introduces SharedClassCache feature to the appsody java stack (which uses openj9). With this enhancement the application image can be built with class data sharing feature turned on and cache is generated as part of build process and the application (JVM) startup is improved when its running.

As the default size for `scc` is 300mb, we are limiting it to 80MB.

Signed-off-by: bharathappali <bharath.appali@gmail.com>


ebullient:
I would like to wait on this one. It's a great idea, but I'd like to do it after we bring in better layering (care of #559)

ebullient:
@bharathappali -- Care to try this one again? If you're in there, it would be nice to switch to the adoptopenjdk openj9 ubi image at the same time...

bharathappali:
@ebullient Will update my PR.

incubator/java-spring-boot2/image/project/Dockerfile
```diff
@@ -62,6 +65,10 @@ COPY --chown=java_user:java_group --from=compile ${DEPENDENCY}/BOOT-INF/classes
 
 USER java_user
 
+COPY --from=compile /tmp/scc/ /tmp/scc/
```
ebullient: COPY --chown=java_user:java_group --from=compile /tmp/scc /tmp/scc

bharathappali: Updated the PR, Thanks @ebullient

bharathappali:
@ebullient I couldn't find ubi image in AdoptOpenJDK official docker image (As we are using `adoptopenjdk:8-jdk-openj9` which is an official docker image), Can we change it to `adoptopenjdk/openjdk8-openj9:ubi-minimal` ?

bharathappali:
I might be wrong please correct me if i'm, Can we use a `JRE` image instead of a `JDK` image ? as i see we are running the java class / jar in `start.sh` and most of the applications doesn't use the mechanism of compiling in runtime, So i was thinking if we could change the base image as `JRE` (To reduce size of image as 80 MB is getting added as part of SCC)

ebullient:
adoptopenjdk/openjdk8-openj9:ubi-minimal is preferred, yes. @BarDweller .. what do you think re: JDK vs. JRE?

ymanton:
Hi, is an 80MB SCC necessary? I would recommend that you size the SCC as small as possible, aiming for 80-90% utilization.

If this image is going to be used as a base for another image and you expect more data to be added to the SCC the dependent image can create a new SCC layer in the existing cache with its own size and add its data to the new layer. If done this way you don't need to ship an SCC with a lot of free space in this image.

(Multi-layer SCCs are a new feature in OpenJ9 and were created for scenarios such as this.)

ebullient:
I would keep it simple (@ the app layer) for now. @BarDweller and I are wrestling with some concerns about how the server is started, and whether or not we should be doing something a bit different for that warm up.

bharathappali:
> is an 80MB SCC necessary?
I have found ~45MB as 96% utilization of SCC (via `printStats` option) for the sample app of the stack. So made it 80MB adding 35MB more space.

> If this image is going to be used as a base for another image
Thanks for suggesting this. wrt to multi layer SCC i haven't considered this scenario thinking the docker image we generate will be used as the final image which will be deployed (general use case)

bharathappali:
@ymanton I'm thinking as the image which is getting generated as part of `appsody build/deploy` will likely to be the final image and we are not adding SCC to java stack image (builder image) so we can be limiting the SCC to be 45~50 MB.

neeraj-laad:
@bharathappali just wondered if there was active work on this one?

bharathappali:
@neeraj-laad Sorry for late response. I was pulled into other work earlier. I have updated the PR and was waiting on @ymanton inputs for the scc size and current flow of creating scc

ymanton:
Sorry, didn't realize you were waiting for feedback from me.

If 45MB is 96% utilized and this is the top-most image (from a Java POV) then yes a 45-50MB SCC makes sense.

bharathappali:
@ebullient Can i get your review on the updated PR ?

neeraj-laad:
@ebullient @BarDweller @bharathappali 

Just trying to understand if we still want to get this PR in? It has been inactive fora while and I was wondering if we should close this, and re-open when we are ready?

ebullient (approved):

bharathappali:
@ebullient Thanks for approving the changes, It's been a long time I haven't tracked the changes in this codebase. I might need a couple of days to rebase and push the changes as I'm currently occupied with other work. Thanks in advance.

scottkurz:
@bharathappali  I just want to level-set too and note that this might not ever get published within a new release, (though this repo has not been archived).

Please see this note:   https://appsody.dev/blogs/DevelopmentEnded

Thank you and sorry about that.

bharathappali:
@scottkurz Thanks for your reply, Sorry haven't checked appsody for a while so was unaware that its development ended.

PR #579, opened 2020-01-03
[python-flask] Run application as non-root

### LINKED ISSUE #351: The python-flask stack runs flask as root - we should switch to non-root
The current Dockerfile-stack runs flask (and hence the user application) as root. We should be able to modify this create a user and run under those privileges.

henrynash:
The current python-flask stack runs the application as root - this is not good practice. This update modifies both the Dockerfile-stack and Dockerfile so that the flask application is run as non-root (when either being controlled by appsody or as a final application image)

Fixes #351


incubator/python-flask/image/Dockerfile-stack
```diff
@@ -24,11 +31,10 @@ ENV APPSODY_TEST="python -m unittest discover -s test -p *.py"
 ENV APPSODY_TEST_ON_CHANGE=$APPSODY_TEST
 ENV APPSODY_TEST_KILL=false
 
-COPY ./LICENSE /licenses/
-COPY ./project /project
-COPY ./config /config
-WORKDIR /project
-RUN pipenv install
+COPY --chown=worker:worker ./LICENSE /licenses/
```
Kamran64: If we're using the chown flag here, please can you add a requirements field to the `stack.yaml`. I.e.
```
.
.
.
default-template: simple
license: Apache-2.0
requirements:
  docker-version: ">= 17.09.0"
```
This will stop users on older versions of Docker hitting errors after they've already started developing.

henrynash: will do, thx

PR #587, opened 2020-01-09
java-openliberty: App startup improvements

awisniew90:
### Checklist:

1. Create server during `appsody stack package` to avoid recreating on each `appsody run`
2. Refactored stack package vs stack run profiles
3. Fixed OL artifact inconsistency between `appsody run` and "host mode"
4. Added go template variables for stack version 
5. Fixed OL download hang during `appsody build` with -B maven option
6. Added shared class cache option by running configure.sh in OL image during `appsody build` (from PR https://github.com/appsody/stacks/pull/596/files)

### DIFF
--- a/incubator/java-openliberty/image/Dockerfile-stack
+++ b/incubator/java-openliberty/image/Dockerfile-stack
@@ -15,7 +15,7 @@ RUN  /project/util/check_version build
 WORKDIR /project/
 RUN mkdir -p /mvn/repository
 RUN mvn -B -Dmaven.repo.local=/mvn/repository -N io.takari:maven:wrapper -Dmaven=$(mvn help:evaluate -Dexpression=maven.version -q -DforceStdout)
-RUN mvn -B -Pstack-package -Dmaven.repo.local=/mvn/repository liberty:install-server liberty:create install dependency:go-offline
+RUN mvn -B -Pstack-image-package -Dmaven.repo.local=/mvn/repository liberty:install-server liberty:create install dependency:go-offline
 RUN chmod -R 777 /opt/ol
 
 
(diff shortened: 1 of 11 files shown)


neeraj-laad:
Requested review from other stack maintainers.

incubator/java-openliberty/templates/default/src/main/liberty/config/server.xml
```diff
@@ -1,6 +1,7 @@
 <server description="Liberty server">
     <featureManager>
-        <feature>microProfile-3.2</feature>
+        <feature>mpHealth-2.0</feature>
+        <feature>mpMetrics-2.0</feature>
```
scottkurz: Two thoughts:

1) This should probably have at least jaxrs-2.1 to enable the JAX-RS resource in the template (which should probably have an IT, though that could be a separate issue for whenever)

2) Though this is just a template it's got the "breaking" characteristic that it removing certain API/features.    Also there's not any particular reason it needs to be merged together with the base stack change.   Maybe then it'd be better to save such a change for a separate PR, minor update.      This way our v0.1.1 change can be described identically for new & existing projects.    (Though this won't always be possible, so not 100% sure what I'm proposing is the right thing. What do you think?)

incubator/java-openliberty/image/Dockerfile-stack
```diff
@@ -34,17 +34,17 @@ ENV APPSODY_USER_RUN_AS_LOCAL=true
 # Set APPSODY_DEV_MODE to get in the habit in case we need it someday
 ENV APPSODY_PREP="export APPSODY_DEV_MODE=prep;  ../validate.sh"
 
-ENV APPSODY_RUN="export APPSODY_DEV_MODE=run; mvn -B -DinstallDirectory=/opt/ol/wlp -Dmaven.repo.local=/mvn/repository liberty:dev"
+ENV APPSODY_RUN="export APPSODY_DEV_MODE=run; mvn -B -DappsDirectory=apps -DinstallDirectory=/opt/ol/wlp -Dmaven.repo.local=/mvn/repository liberty:dev"
```
scottkurz: I know this is in part a workaround for [this](https://github.com/OpenLiberty/ci.maven/issues/718).    Might be cleaner to just add <appsDirectory> config once in POM rather than in each mvn invocation.  Also if it were nicer to put in a template POM rather than the stack POM, then you require each template to know about this .. which argues for putting in the stack POM (following my idea).    Is that a good fit?   Well, that also suggests we could use some more doc on how to navigate through the plugin config in each POM/mode.

scottkurz: I also don't see why we need to include the `-DinstallDirectory=/opt/ol/wlp`.     Can we try collapsing the stack package path and the `appsody run/debug/test` paths back together (but leaving distinct from the host-only path)?   I think we either wrongly reacted to some other bug or are dealing with docker timestamp lag (and if the latter, I'd like to raise this to the ci.maven project).

incubator/java-openliberty/templates/default/src/main/liberty/config/server.xml
awisniew90: I think it would be best to leave microProfile-3.2 as the feature definition here. In a separate PR I can slim this down to mpHealth, mpMetrics, and jaxrs while adding an IT to test the app endpoint.

incubator/java-openliberty/image/Dockerfile-stack
awisniew90: 1. The `appsDirectory` property will need to be set in the child pom. If we set it in the parent pom, it would be set in the stack-package profile which isnt invoked when liberty:dev is called on the child. Since this is a temporary workaround (and only a problem when we are in the docker / appsody run flow), I think we should leave it here rather than muddling with the child pom.  

2. I dont see a clean way to collapse these together. The idea (I believe) would be to have the child pom inherit the `runtimeInstallDirectory` property from the parent. Last time we tested this, I believe the profile definitions were a bit different, but now, the child can't inherit this property for the same reason above^. The stack-package profile is not active when the child pom builds. But even if it were, it would then just expose the potential timestamp bug that leads to uninstalling the runtime. Setting the `installDirectory` property on the child pom tells the LMP that the runtime is _already_ installed rather than telling it a location _to install_ it (or potentially reinstall it). If this is a bug in ci.maven, once fixed we can just remove the -DinstallDirectory property from the maven invocations just like we can remove the -DappsDirectory property once that bug is addressed.

incubator/java-openliberty/templates/default/src/main/liberty/config/server.xml
awisniew90: I re-added the microProfile-3.2 feature and opened https://github.com/appsody/stacks/issues/592 to slim down the feature list and add an IT to test jaxrs.

scottkurz:
@Kamran64  this is ready to review for merge/release.   Since we batched some changes together We tried to bubble up the "release notes" in the initial comment.  If that's a bit too detailed we can take another pass, but would like to show a bit more than just:  `App startup improvements`

PR #594, opened 2020-01-13
Binary war file project template for java-openliberty stack

### LINKED ISSUE #387: Allow appsody to work with existing project binaries
It would be useful if I could use appsody as a mechanism for deploying existing projects that have been pre-built into relevant binaries. (for example *.war, *.ear for java cases).

In these cases, users are not after the local development flow and benefits of using appsody during those stages but are looking to use appsody for deployment management and other benefits.

ajm01:
### Checklist:

- [x ] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stacks-overview.md#stack-structure).

### Contributing a new stack:

### Related Issues:

Fixes #387


ajm01:
Fixes #387

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+        <repository>
+            <id>project.local</id>
+            <name>project</name>
+            <url>file:${project.basedir}/${file.repo.name}</url>
+        </repository>
+    </repositories>
+
+    <dependencies>
+        <!-- Aggregate of MicroProfile APIs -->
+        <dependency>
```
scottkurz: Did we really want to include this?  We're not building anything right

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+                <executions>
+                    <execution>
+                        <id>copy-war-dependency</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>copy-dependencies</goal>
+                        </goals>
+                        <configuration>
+                            <!-- put war dependency file in target after dependency resolution -->
+                            <outputDirectory>${project.build.directory}</outputDirectory>
```
scottkurz: So we're keeping the WAR name the same?

incubator/java-openliberty/templates/binary/src/main/liberty/config/server.xml
```diff
@@ -0,0 +1,11 @@
+<server description="Liberty server">
+    <featureManager>
+        <feature>microProfile-3.0</feature>
+    </featureManager>
+    <quickStartSecurity userName="admin" userPassword="adminpwd"/>
+    <keyStore id="defaultKeyStore" location="key.jks" type="jks" password="mpKeystore"/>
+    <httpEndpoint host="*" httpPort="${default.http.port}" 
+        httpsPort="${default.https.port}" id="defaultHttpEndpoint"/>
+    
+    <webApplication location="starter-app.war" contextRoot="/"/>
```
scottkurz: Are we copying into this WAR name or using the original migrated app WAR name?

incubator/java-openliberty/image/project/Dockerfile
```diff
@@ -18,7 +18,7 @@ COPY ./validate.sh /project/user-app/validate.sh
 RUN cd /project/user-app && ./validate.sh
 
 # Copy and build the application source, and run unit tests
-COPY ./user-app/src /project/user-app/src
+COPY ./user-app /project/user-app
 RUN cd /project/user-app && mvn liberty:create package
 
 RUN cd /project/user-app/target/liberty/wlp/usr/servers && \
```
scottkurz: I think we should go further and separately copy the libs and resources, e.g. something like:
```
 # Stage 2: Package Open Liberty image
 FROM open-liberty:kernel-java8-openj9

 # 2a) shared resources change the least-frequently  (* - see note below)
 COPY --chown=1001:0 --from=0 /shared/resources /opt/ol/wlp/usr/shared/

 # 2b) next copy shared library
 COPY --chown=1001:0 --from=0 /config/libs/ /config/

 # 2c) Server config, bootstrap.properties, and everything else
 COPY --chown=1001:0 --from=0 /config/ /config/

 # 2d) Changes to the application binary
 COPY --chown=1001:0 --from=0 /project/user-app/target/*.war /config/apps
```

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <!--required parent POM-->
+        <groupId>dev.appsody</groupId>
+        <artifactId>java-openliberty</artifactId>
+        <version>[0.1, 0.2)</version>
```
scottkurz: should use Go template var

ajm01: done

incubator/java-openliberty/image/project/Dockerfile
ajm01: done

incubator/java-openliberty/templates/binary/pom.xml
ajm01: true, removed

incubator/java-openliberty/templates/binary/src/main/liberty/config/server.xml
ajm01: we will preserve the original war name - the original war name will be the name of the app installed in the final image and will be the app name that should be reflected into any server.xml included in the project instance.

scottkurz: so this app in the template server.xml isn't helping.  Maybe just add **src/main/liberty/config/.gitignore** to preserve the location but don't add server.xml?

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,156 @@
+<?xml version='1.0' encoding='utf-8'?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
```
scottkurz: Hmm...arguable, but maybe we should drop the parent altogether.  Without the local dev mode, I'm not sure it's buying us anything.

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+        <!--required parent POM-->
+        <groupId>dev.appsody</groupId>
+        <artifactId>java-openliberty</artifactId>
+        <version>{{.stack.parentpomrange}}</version>
+    </parent>
+
+    <groupId>dev.appsody.starter.java-openliberty</groupId>
+    <artifactId>tamigration</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>war</packaging>
```
scottkurz: Not sure what type of packaging.. but since "war" is requiring us to disable the war plugin...  and since "jar" is more the default...maybe "jar" packaging?     

One could also argue for "pom" packaging.  But it's not crazy to think someone could want to package a jar with all this.  

OTOH, it is a bit odd to have one stack containing multiple packaging types.  But "binary" is going to be a slightly odd-fit no matter what.   I still vote for .jar

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+                    <failOnMissingWebXml>false</failOnMissingWebXml>
+                    <packagingExcludes>pom.xml</packagingExcludes>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+            <!-- Enable liberty-maven plugin -->
+            <plugin>
+                <groupId>io.openliberty.tools</groupId>
+                <artifactId>liberty-maven-plugin</artifactId>
+                <configuration>
```
scottkurz: Why is there config at all?   This sort of opens the door to asking if we do need a parent to constrain the version.   But if we're using nothing more from this locally-installed server than the directory layout...maybe we should just always use the latest and assume it'll never change?

incubator/java-openliberty/image/project/Dockerfile
```diff
@@ -13,22 +13,44 @@ COPY ./util /project/util
...
+#     but can't assume config/lib exists - copy from previous stage to a tmp holding place and test
+COPY --chown=1001:0 --from=0 /configlibdir/ /tmp/cfglib
+RUN if [ "$(ls -A /tmp/cfglib/lib)" ]; then cp -R /tmp/cfglib/lib /config && chmod -R 777 /tmp/cfglib && rm -rf /tmp/cfglib; fi
+
+# 2c) Server config, bootstrap.properties, and everything else
+COPY --chown=1001:1001 --from=0 /config/ /config/
+
+
+# 2d) Changes to the application binary
 COPY --chown=1001:0 --from=0 /project/user-app/target/*.war /config/apps
```
scottkurz: Noticing `appsody stack validate` fails w/ init'd template, since there's no .war present...  Not sure we won't hit a next problem but to get past this one, it would certainly be valid to have no WAR present.   That is.. guard this with a test.

incubator/java-openliberty/templates/binary/pom.xml
scottkurz: OK..one point I hadn't thought was the idea that the stack author could "blacklist" certain dependencies.    This would prevent them from getting built into the Docker image.   Maybe we should say that's valid then, reintroduce the parent, and adjust validate.sh to not blow up for our binary template?

incubator/java-openliberty/image/project/Dockerfile
```diff
@@ -13,22 +13,45 @@ COPY ./util /project/util
...
+#     but can't assume config/lib exists - copy from previous stage to a tmp holding place and test
+COPY --chown=1001:0 --from=0 /configlibdir/ /tmp/cfglib
+RUN if [ "$(ls -A /tmp/cfglib/lib)" ]; then cp -R /tmp/cfglib/lib /config && chmod -R 777 /tmp/cfglib && rm -rf /tmp/cfglib; fi
+
+# 2c) Server config, bootstrap.properties, and everything else
+COPY --chown=1001:1001 --from=0 /config/ /config/
+
+
+# 2d) Changes to the application binary
+COPY --chown=1001:0 --from=0 /project/user-app/target/* /tmp/wars/
```
scottkurz: This should only copy `/project/user-app/target/*.war` I think?

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+                    </execution>
+                    <!-- example of how to move a resource adapter dependency
+                    <execution>
+                        <id>copy-resource-adapter-artifacts</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>copy-dependencies</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/liberty/wlp/usr/servers/shared/resources</outputDirectory>
```
scottkurz: This is just a comment but I think you meant: 
```
 <outputDirectory>${project.build.directory}/liberty/wlp/usr/shared/resources</outputDirectory>
```

scottkurz: (No "servers" in that path).

incubator/java-openliberty/image/project/Dockerfile
```diff
@@ -13,22 +13,45 @@ COPY ./util /project/util
...
 
 # Step 2: Package Open Liberty image
 FROM open-liberty:kernel-java8-openj9
 
-COPY --chown=1001:0 --from=0 /config/ /config/
-COPY --chown=1001:0 --from=0 /project/user-app/target/*.war /config/apps
+#2a) copy any resources 
+#    but, can't assume shared/resources contains files - copy from previous stage to a tmp holding place and test
+COPY --chown=1001:1001 --from=0 /shared /tmp/shared
+RUN if [ -d /tmp/shared/resources ]; then cp -R /tmp/shared/resources /opt/ol/wlp/usr/shared/ && chmod -R 777 /tmp/shared && rm -rf /tmp/shared; fi
```
scottkurz: In line 34 above we copied everything from **usr/shared**... but here we're only copying **shared/resources**.  Maybe we should just copy everything, since from [here](https://www.ibm.com/support/knowledgecenter/SS7K4U_liberty/com.ibm.websphere.wlp.zseries.doc/ae/rwlp_dirs.html) we see also there are **shared.config.dir** and other special directories.

incubator/java-openliberty/templates/binary/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+                            <overWriteReleases>false</overWriteReleases>
+                            <overWriteSnapshots>true</overWriteSnapshots>
+                            <excludeTransitive>true</excludeTransitive>
```
scottkurz: How did we come up with all these config parm values?  

I think the most helpful one might simply be:
```
      <includeGroupIds>mysql</includeGroupIds>
```

Besides that maybe a more helpful thing would be a link to the doc?  

One gotcha is if you're iterating through a 1.0 release testing this out (instead of using a snapshot), then overwrite false can prevent changes.   OK, maybe you should know better... I'd just as soon step aside altogether if we don't have a good reason for suggesting these values).

incubator/java-openliberty/image/Dockerfile-stack
```diff
@@ -34,7 +34,7 @@ ENV APPSODY_USER_RUN_AS_LOCAL=true
 # Set APPSODY_DEV_MODE to get in the habit in case we need it someday
 ENV APPSODY_PREP="export APPSODY_DEV_MODE=prep;  ../validate.sh"
 
-ENV APPSODY_RUN="export APPSODY_DEV_MODE=run; mvn -B -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository liberty:dev"
+ENV APPSODY_RUN="export APPSODY_DEV_MODE=run; /project/invoke-appsody.sh"
```
scottkurz: You'd need to do this for test/debug also.

I have a better idea..let's just move this function into validate.sh, (where we'll continue to test APPSODY_DEV_MODE so we don't run it in `appsody build`), since that runs in PREP.

incubator/java-openliberty/image/project/invoke-appsody.sh
```diff
@@ -0,0 +1,34 @@
...
+then
+	BIN_TEMPLATE=1
+fi
+
+case $APPSODY_DEV_MODE in
+	run)
+		#if [ -e /project/user-app/.appsody-nolocal ]
+		if [ $BIN_TEMPLATE = 1 ]
+		then
+			echo this appsody run not supported when using binary template. Press ctl-C and/or stop the current container.
```
scottkurz: Nit picking... there's more than one way to do this but I don't recall seeing `ctl-C`.  Could we maybe use `Ctrl+C` or something more common.

incubator/java-openliberty/image/project/invoke-appsody.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+if [ -e /project/user-app/.appsody-nolocal ]
```
scottkurz: For now let's call it `.appsody-binary`.   If needed we could broaden this and rename to something like this..For now it might be more obvious to tie it to binary.

incubator/java-openliberty/image/project/invoke-appsody.sh
```diff
@@ -0,0 +1,34 @@
...
+then
+	BIN_TEMPLATE=1
+fi
+
+case $APPSODY_DEV_MODE in
+	run)
+		#if [ -e /project/user-app/.appsody-nolocal ]
+		if [ $BIN_TEMPLATE = 1 ]
```
scottkurz: This produces an error message if BIN_TEMPATE isn't set.  

Maybe a better test would be:

`if [[ -f <file> ]]`

incubator/java-openliberty/templates/binary/pom.xml
ajm01: ok, updated to jar packaging

incubator/java-openliberty/templates/binary/pom.xml
ajm01: parent pom stanza re-introduced, no longer skipping the check for it if binary template

incubator/java-openliberty/image/project/Dockerfile
ajm01: now guarded with a test

incubator/java-openliberty/image/project/Dockerfile
ajm01: could be an ear file as well, i don't think it is possible to do a conditional docker copy...so need to copy everything to a temp area and then look specifically for war and/or ear files

incubator/java-openliberty/templates/binary/pom.xml
ajm01: yes, will fix these commented out stanzas

incubator/java-openliberty/image/project/Dockerfile
ajm01: done

incubator/java-openliberty/templates/binary/pom.xml
ajm01: i have updated the config of these stanzas to reflect the need for <includeGroupIds> ... i also detail <includeArtifactIds> as well

incubator/java-openliberty/image/project/invoke-appsody.sh
ajm01: done

incubator/java-openliberty/image/project/invoke-appsody.sh
ajm01: eliminated this part of the message by moving this all to validate.sh

incubator/java-openliberty/image/project/invoke-appsody.sh
ajm01: this part is removed now, no longer needed. function moved to validate.sh

ajm01:
closing as per discussion - will use new pr to introduce just the dockerfile updates

PR #617, opened 2020-01-23
java-openliberty: update validate script

awisniew90:
### Checklist:

- [ x] Updated the stack version in `stack.yaml`

Currently we are getting parent pom properties in the validate script (groupid, artifactid, version) via a call to the maven exec plugin. This takes time and can be replaced by building a DOM on the pom (which we are already doing further down in the script. 

This change cuts off about 5 seconds for `appsody run`.

### DIFF
--- a/incubator/java-openliberty/image/project/validate.sh
+++ b/incubator/java-openliberty/image/project/validate.sh
@@ -22,16 +22,6 @@ if [ ! -z "$APPSODY_DEV_MODE" ]; then
     M2_LOCAL_REPO="-Dmaven.repo.local=/mvn/repository"
 fi
 
-
-# Get parent pom information (../pom.xml)
-args='export PARENT_GROUP_ID=${project.groupId}; export PARENT_ARTIFACT_ID=${project.artifactId}; export PARENT_VERSION=${project.version}
-export LIBERTY_GROUP_ID=${liberty.groupId}; export LIBERTY_ARTIFACT_ID=${liberty.artifactId}; export LIBERTY_VERSION=${version.openliberty-runtime}'
-eval $(mvn -q -Dexec.executable=echo $M2_LOCAL_REPO -Dexec.args="${args}" --non-recursive -f ../pom.xml exec:exec 2>/dev/null)
-
-# Install parent pom
-echo "Installing parent ${PARENT_GROUP_ID}:${PARENT_ARTIFACT_ID}:${PARENT_VERSION}"
-mvn install  $M2_LOCAL_REPO -Denforcer.skip=true -f ../pom.xml
-
 # Get parent pom information
 a_groupId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:groupId" /project/pom.xml)
 a_artifactId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:artifactId" /project/pom.xml)
@@ -40,6 +30,12 @@ p_groupId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x
 p_artifactId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:parent/x:artifactId" pom.xml)
 p_version_range=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:parent/x:version" pom.xml)
 
+# Install parent pom
+echo "Installing parent ${a_groupId}:${a_artifactId}:${a_version}"
+mvn install $M2_LOCAL_REPO -Denforcer.skip=true -f ../pom.xml
+
+
+
 # Check child pom for required parent project
 if [ "${p_groupId}" != "${a_groupId}" ] || [ "${p_artifactId}" != "${a_artifactId}" ]; then
   echo "Project pom.xml is missing the required parent:
@@ -47,8 +43,7 @@ if [ "${p_groupId}" != "${a_groupId}" ] || [ "${p_artifactId}" != "${a_artifactI
   <parent>
     <groupId>${a_groupId}</groupId>
     <artifactId>${a_artifactId}</artifactId>
-    <version>${a_range}</version>
-    <relativePath/>
+    ...
   </parent>
   "
   exit 1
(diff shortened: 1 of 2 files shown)


incubator/java-openliberty/image/project/validate.sh
```diff
@@ -12,15 +12,6 @@ if [ ! -f ./pom.xml ]; then
   exit 1   
 fi
 
-# Get parent pom information (../pom.xml)
-args='export PARENT_GROUP_ID=${project.groupId}; export PARENT_ARTIFACT_ID=${project.artifactId}; export PARENT_VERSION=${project.version}
-export LIBERTY_GROUP_ID=${liberty.groupId}; export LIBERTY_ARTIFACT_ID=${liberty.artifactId}; export LIBERTY_VERSION=${version.openliberty-runtime}'
```
scottkurz: How does removing this affect the tests later on in lines 69-73?

By using DOM naively we'd have a hard time taking into account profiles I'd think... are we sure we don't care?   I guess at a high level what were we actually checking before?

incubator/java-openliberty/image/project/validate.sh
```diff
@@ -29,14 +20,20 @@ p_groupId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x
...
+
+
 # Check child pom for required parent project
 if [ "${p_groupId}" != "${a_groupId}" ] || [ "${p_artifactId}" != "${a_artifactId}" ]; then
   echo "Project pom.xml is missing the required parent:
 
   <parent>
     <groupId>${a_groupId}</groupId>
     <artifactId>${a_artifactId}</artifactId>
-    <version>${a_range}</version>
```
scottkurz: This is just something we hadn't realized was wrong?  (I don't see it set anywhere)?   I guess it's just an error msg though.  Good catch.

incubator/java-openliberty/image/project/validate.sh
awisniew90: We're no longer doing any of the checks from line 69 on (we skip those and `exit 0` earlier). We also pushed the liberty artifact configuration into the parent pom, so the child no longer has this anyways, making the check obsolete.

incubator/java-openliberty/image/project/validate.sh
awisniew90: Right, probably a typo. The parent doesnt have a range for it's version, so we can only recommend a set version number.

incubator/java-openliberty/image/project/validate.sh
scottkurz: I didn't realize that.  

What if we removed all that stuff and replaced with a check that the `version.liberty-maven-plugin` property hasn't been overridden?     This too would be more thorough a check if we checked from a maven context than just statically reading the pom.xml (since, assuming we add something like MVN_ARGS.. we could "cheat").  But it's maybe a good compromise with what this originally set out to do, and better than leaving this stuff in there?

awisniew90: Good idea. I can add that in.

incubator/java-openliberty/image/project/validate.sh
```diff
@@ -12,15 +12,6 @@ if [ ! -f ./pom.xml ]; then
...
 
-# Get parent pom information (../pom.xml)
-args='export PARENT_GROUP_ID=${project.groupId}; export PARENT_ARTIFACT_ID=${project.artifactId}; export PARENT_VERSION=${project.version}
-export LIBERTY_GROUP_ID=${liberty.groupId}; export LIBERTY_ARTIFACT_ID=${liberty.artifactId}; export LIBERTY_VERSION=${version.openliberty-runtime}'
-eval $(mvn -q -Dexec.executable=echo -Dmaven.repo.local=/mvn/repository -Dexec.args="${args}" --non-recursive -f ../pom.xml exec:exec 2>/dev/null)
-
-# Install parent pom
-echo "Installing parent ${PARENT_GROUP_ID}:${PARENT_ARTIFACT_ID}:${PARENT_VERSION}"
-mvn install -Dmaven.repo.local=/mvn/repository -Denforcer.skip=true -f ../pom.xml
-
```
scottkurz: SK1:  OK, I get it and I like this part of the change !

incubator/java-openliberty/image/project/validate.sh
```diff
@@ -29,14 +20,20 @@ p_groupId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x
...
+
 # Check child pom for required parent project
 if [ "${p_groupId}" != "${a_groupId}" ] || [ "${p_artifactId}" != "${a_artifactId}" ]; then
   echo "Project pom.xml is missing the required parent:
 
   <parent>
     <groupId>${a_groupId}</groupId>
     <artifactId>${a_artifactId}</artifactId>
-    <version>${a_range}</version>
+    <version>${a_version}</version>
```
scottkurz: SK1: I think the "version" part of this error message is more confusing than helpful.   What if we just said:
```
  <parent>
    <groupId>${a_groupId}</groupId>
    <artifactId>${a_artifactId}</artifactId>
     ...
```

or even condensed it to a single-line message:   `incorrect parent groupId:artifactId, expecting aaaa:bbbb, found xxxx:yyy`

scottkurz: SK1: We do the check_version below..let's leave the versioning to that.

incubator/java-openliberty/image/project/validate.sh
```diff
@@ -54,39 +51,30 @@ the parent version in pom.xml, and test your changes.
...
-# Skip check below, not sure if we should fix or just remove.   It doesn't account for the fact that
-# the stack now uses pluginManagement, so we could fix to allow this as an acceptable usage too... but what if Jane introduces a profile?
-exit 0
 
-# Check child pom for required liberty version, groupID and artifactId
-l_groupId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:build/x:plugins/x:plugin[x:artifactId='liberty-maven-plugin']/x:configuration/x:assemblyArtifact/x:groupId" pom.xml)
-l_artifactId=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:build/x:plugins/x:plugin[x:artifactId='liberty-maven-plugin']/x:configuration/x:assemblyArtifact/x:artifactId" pom.xml)
-l_version=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:build/x:plugins/x:plugin[x:artifactId='liberty-maven-plugin']/x:configuration/x:assemblyArtifact/x:version" pom.xml)
+# Check that the child pom has not overridden the Open Liberty version
+child_ol_version_property=$(xmlstarlet sel -T -N x="http://maven.apache.org/POM/4.0.0" -t -v "/x:project/x:properties/x:version.openliberty-runtime" pom.xml)
```
scottkurz: SK1:  I like the direction you headed in, and I get it now.

Perhaps all we need to look at for now is whether there is a **runtimeArtifact** at all?   I don't see that allowing  the element **runtimeArtifact** but then requiring it to have empty child elements is OK..Jane has gone astray.    

Note there are several other [methods](https://github.com/OpenLiberty/ci.maven/blob/master/docs/installation-configuration.md) for installing liberty that you should also NOT be doing, that we COULD check for.

I wish we could wait until we've gotten the install down and read  `lib/versions/openliberty.properties` to see what version's installed.  Of course...what if the config later points us somewhere else?   IDK.   

I don't think this is adding that much value over the enforcer OL version property check...but it does seem like it's the most likely thing for Jane to try to do that shouldn't be allowed (by copying/pasting from the parent POM and tweaking the version), so I'm OK with it.

awisniew90: Good points.  The intent with the "empty" checks was to make sure those *runtimeArtifact* elements weren't set to begin with. I hadn't thought that Jane could define them but keep them empty which would be strange. 

Yea I think the general idea was to cover the gap of Jane accidentally (or misguidedly) adding her own runtime artifact and not realizing that is not ok to do. This would flag that scenario, although as you said, there are other methods for installing liberty not covered with this. 

Since you are ok with this, lets leave it in and if we want we can revisit how to make this a bit more robust. Or leave it to Champ to do as he chooses.

incubator/java-openliberty/image/project/validate.sh
awisniew90: Done

incubator/java-openliberty/image/project/validate.sh
```diff
@@ -65,39 +60,30 @@ the parent version in pom.xml, and test your changes.
   <parent>
     <groupId>${a_groupId}</groupId>
     <artifactId>${a_artifactId}</artifactId>
-    <version>${a_range}</version>
+    <version>${a_version}</version>
```
scottkurz: p_version_range

scottkurz:
Closing, rolling into #650

PR #634, opened 2020-02-04
dockerfile updates to support a pre-existing war/ear file project

ajm01:
### Checklist:

- [x ] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stacks-overview.md#stack-structure).

### Contributing a new stack:

### DIFF
--- a/incubator/java-openliberty/image/project/Dockerfile
+++ b/incubator/java-openliberty/image/project/Dockerfile
@@ -2,7 +2,7 @@
 FROM adoptopenjdk/openjdk8-openj9
 
 RUN apt-get update && \
-    apt-get install -y maven unzip wget xmlstarlet
+	apt-get install -y maven unzip wget xmlstarlet
 
 # Copy and build the dev.appsody:java-openliberty parent pom
 COPY ./pom.xml /project/pom.xml
@@ -18,7 +18,7 @@ RUN cd /project/user-app && \
 
 # Copy and run a simple version check
 COPY ./util /project/util
-RUN  /project/util/check_version build
+RUN	 /project/util/check_version build
 
 # Copy the validate.sh script and application pom.xml
 COPY ./validate.sh /project/user-app/validate.sh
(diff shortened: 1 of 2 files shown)


incubator/java-openliberty/image/project/Dockerfile
```diff
@@ -28,21 +28,41 @@ COPY ./user-app/pom.xml /project/user-app/pom.xml
 RUN cd /project/user-app && ./validate.sh
 
 # Copy the rest of the application source
-COPY ./user-app/src /project/user-app/src
+COPY ./user-app /project/user-app
```
scottkurz: Since we're copying everything, which might include a /target (even a /target/wlp) from running in the host.... we should probably change line 36 below to do a clean:
`mvn -B liberty:create package`

incubator/java-openliberty/stack.yaml
```diff
@@ -23,3 +23,4 @@ requirements:
 templating-data:
     libertyVersion: '19.0.0.12'
     parentpomrange: '[0.1, 0.2)'
+
```
scottkurz: extra line

ajm01: removed extra line

incubator/java-openliberty/image/project/Dockerfile
ajm01: added a clean step

scottkurz: OK, that was my bad idea... since it eliminates the benefit of having done the `install-server` above.

Even if there is no overwriting, there is still the possibility you get different results depending on what you happened to do on the host or not which would be confusing, so we should do this.

What we should use is **.dockerignore** to exclude target.   Actually it already does :)  so just remove the clean then.
----

Also, I know this is a bigger change so am open to doing it later, but...   

We don't need the whole 150MB         <liberty.artifactId>openliberty-runtime</liberty.artifactId>   for this... we just need the kernel.   That's be a nice way to slim down a bit.. we can use another profile.

ajm01:
Hi @Kamran64 - this PR is ready to be merged, thanks

PR #635, opened 2020-02-05
java-openliberty: update LMP version to 3.2

### LINKED ISSUE #608: java-openliberty:  Pick up LMP V3.2 and delegate to liberty-maven-plugin to watch pom.xml changes
As commented [here](https://github.com/appsody/stacks/blob/deb09f9d8a404b6105a4ee4e56ebce6d9c5e6614/incubator/java-openliberty/image/Dockerfile-stack#L29-L30) we are using the appsody watcher to watch pom.xml as a workaround to v3.1 liberty-maven-plugin limitations.

Once https://github.com/OpenLiberty/ci.maven/pull/676 is available in a release, we should move our appsody stack to that version of the plugin, and remove this special case exception from our appsody watcher.

awisniew90:
### Checklist:

- [ x] Read the [Code of Conduct](https://github.com/appsody/website/blob/master/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md).

* Update the Liberty Maven Plugin to version 3.2
* Add requirement in enforcer plugin to use LMP version 3.2

### Related Issues:
Fixes #608 

### Release Notes:
* Updates the Liberty Maven plugin version to 3.2

### DIFF
--- a/incubator/java-openliberty/image/Dockerfile-stack
+++ b/incubator/java-openliberty/image/Dockerfile-stack
@@ -26,8 +26,7 @@ ENV APPSODY_MOUNTS="~/.m2/repository:/mvn/repository;.:/project/user-app"
 ENV APPSODY_WATCH_DIR=/project/user-app
 ENV APPSODY_WATCH_IGNORE_DIR=/project/user-app/target
 
-# Work around v3.1 issues with dev mode reacting to pom.xml change but should eventually not require this.
-ENV APPSODY_WATCH_REGEX="^pom\.xml$"
+ENV APPSODY_WATCH_REGEX=""
 
 ENV APPSODY_USER_RUN_AS_LOCAL=true
 
@@ -35,18 +34,13 @@ ENV APPSODY_USER_RUN_AS_LOCAL=true
 ENV APPSODY_PREP="export APPSODY_DEV_MODE=prep;  ../validate.sh"
 
 ENV APPSODY_RUN="export APPSODY_DEV_MODE=run; mvn -B -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
-# See ENV APPSODY_WATCH_REGEX comment
-ENV APPSODY_RUN_ON_CHANGE="export APPSODY_DEV_MODE=run; ../validate.sh && mvn -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
-ENV APPSODY_RUN_KILL=true
+ENV APPSODY_RUN_ON_CHANGE=""
 
 ENV APPSODY_DEBUG="export APPSODY_DEV_MODE=debug; mvn -B -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
-# See ENV APPSODY_WATCH_REGEX comment
-ENV APPSODY_DEBUG_ON_CHANGE="export APPSODY_DEV_MODE=debug; ../validate.sh && mvn -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
-ENV APPSODY_DEBUG_KILL=true
+ENV APPSODY_DEBUG_ON_CHANGE=""
 
 ENV APPSODY_TEST="export APPSODY_DEV_MODE=test; exec mvn -B -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository clean package liberty:create liberty:install-feature liberty:start liberty:deploy failsafe:integration-test@it-exec liberty:stop failsafe:verify"
 ENV APPSODY_TEST_ON_CHANGE=""
-ENV APPSODY_TEST_KILL=true
 
 ENV PORT=9080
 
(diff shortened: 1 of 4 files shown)


incubator/java-openliberty/image/project/pom.xml
```diff
@@ -54,31 +54,43 @@
...
+				<artifactId>maven-enforcer-plugin</artifactId>
+				<version>3.0.0-M3</version>
+				<executions>
+					<execution>
+						<id>default-cli</id>
+						<goals>
+							<goal>enforce</goal>
+						</goals>
+						<configuration>
+							<rules>
```
scottkurz: Some thoughts:
1. We should err on side of less restriction.  Let Champ be more restrictive if he wants and let this be a bit of an example as to how, but unless we're helping our direct stack users somehow we should do less rather than more.
2. We could just check/enforce the plugin version property but allow Jane to configure the whole plugin if she's motivated enough to do so?      The fact that the config is complicated isn't that bad .. but maybe since I'm really on the fence it's swaying me over to not worrying about the plugin (but just the property).   
3. Maybe it'd help to ban the wasdev plugin v2.x?  In case someone copied/pasted some config from the internet?
4.  enforcing the OL version is good , since it makes it easier to align with the `appsody build`.

incubator/java-openliberty/stack.yaml
```diff
@@ -22,4 +22,5 @@ requirements:
     appsody-version: ">= 0.5.0"
 templating-data:
     libertyVersion: '19.0.0.12'
+    libertyMavenPluginVersion: '3.2'
```
scottkurz: So this only shows up in a single place in the project?   A project POM would be enough to avoid duplication.   I guess you're viewing this as documentation, or bubbling up into one place.   But for something so POM-specific anyway ( a plugin version), maybe a parent POM is better?   Not totally sure....have I made you think more?

awisniew90: My thinking with this was to keep these values one place, but I agree, this might be over complicating things. I can move this to just be defined in the parent pom.

incubator/java-openliberty/image/project/pom.xml
awisniew90: For this particular move (v3.1 -> v3.2), I think we want to be more restrictive since we are modifying our stack to no longer work properly with v3.1. (watching for changes in the pom.xml). Typically, id assume that if we upgraded our stack to use a newer LMP version, it would be to fix a potential bug or add some additional enhancement. I agree it would be good to allow Jane to use any version _above_ what is defined in the stack, but not sure if the enforcer plugin can be configured that way.

incubator/java-openliberty/image/Dockerfile-stack
```diff
@@ -26,22 +26,19 @@ ENV APPSODY_MOUNTS="~/.m2/repository:/mvn/repository;.:/project/user-app"
...
 ENV APPSODY_USER_RUN_AS_LOCAL=true
 
 # Set APPSODY_DEV_MODE to get in the habit in case we need it someday
 ENV APPSODY_PREP="export APPSODY_DEV_MODE=prep;  ../validate.sh"
 
 ENV APPSODY_RUN="export APPSODY_DEV_MODE=run; mvn -B -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
-# See ENV APPSODY_WATCH_REGEX comment
-ENV APPSODY_RUN_ON_CHANGE="export APPSODY_DEV_MODE=run; ../validate.sh && mvn -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
+ENV APPSODY_RUN_ON_CHANGE=""
 ENV APPSODY_RUN_KILL=true
```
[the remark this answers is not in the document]
awisniew90: 👍

scottkurz:
@Kamran64 ready to go..thanks.

scottkurz:
Slowing this down..this is actually a breaking change.  Closing for now.

scottkurz:
Typically the pom version would not be in the user-app project, and so this is not in fact breaking.  So re-opening.

incubator/java-openliberty/image/Dockerfile-stack
```diff
@@ -26,27 +26,21 @@ ENV APPSODY_MOUNTS="~/.m2/repository:/mvn/repository;.:/project/user-app"
...
 ENV APPSODY_USER_RUN_AS_LOCAL=true
 
 # Set APPSODY_DEV_MODE to get in the habit in case we need it someday
 ENV APPSODY_PREP="export APPSODY_DEV_MODE=prep;  ../validate.sh"
 
 ENV APPSODY_RUN="export APPSODY_DEV_MODE=run; mvn -B -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
-# See ENV APPSODY_WATCH_REGEX comment
-ENV APPSODY_RUN_ON_CHANGE="export APPSODY_DEV_MODE=run; ../validate.sh && mvn -Pstack-image-run -DappsDirectory=apps -Dmaven.repo.local=/mvn/repository package liberty:dev"
-ENV APPSODY_RUN_KILL=true
+ENV APPSODY_RUN_ON_CHANGE=""
```
Kamran64: Can we remove this and `APPSODY_DEBUG_ON_CHANGE` since they're not being used?

Kamran64:
@awisniew90 - I discussed this PR with @scottkurz and we can remove all of the `WATCH` variables now since this stack no longer uses the appsody watcher. I've added that in the latest commit.

PR #650, opened 2020-02-09
Changes for 0.2.0

### LINKED ISSUE #259: Appsody Liberty stacks should be based on UBI SFJ images
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Actual behaviour**
What is the actual behaviour.

**Environment Details (please complete the following information):**

**If applicable please specify:**
 - CLI version:
 - Stack you are using:

**Screenshots**

**Additional context**

### LINKED ISSUE #607: java-openliberty: Explore running as non-root user within container
**Is your feature request related to a problem? Please describe.**
 For the java-openliberty stack, within the container one runs as 'root'.

**Describe the solution you'd like**
 In contrast, the SB2 stack see [here](https://github.com/appsody/stacks/blob/master/incubator/java-spring-boot2/image/Dockerfile-stack) runs as **java-user**.  

We should explore doing the same.

**Describe alternatives you've considered**

**Additional context**

### LINKED ISSUE #611: java-openliberty -  Import of default template into Codewind Eclipse IDE gives Java EE facet validation error
**Describe the bug**
After creation of codewind java-openliberty project, you get validation errors such as:
> Context and dependency injection (CDI) 1.2 and Dynamic Web Module 2.5 cannot both be selected.

**To Reproduce**

**Expected behavior**

**Actual behaviour**
What is the actual behaviour.

**Environment Details (please complete the following information):**
 - Codewind 0.7

**If applicable please specify:**
 - CLI version: N/A
 - Stack you are using (including the version):

**Screenshots**

**Additional context**

### LINKED ISSUE #625: Java-openliberty: Keep Open Liberty runtime version congruent between the stack image and build image
Currently, the Open Liberty version of the appsody build image is not managed. We should make sure that this OL image matches the runtime version of OL in our stack image to make sure these two workflows are congruent.

### LINKED ISSUE #639: java-openliberty:  Provide EE APIs to default template?
Today our default template provides the 3.2 aggregate MicroProfile API.

Should we consider providing EE aggregate APIs as well?  (Probably Jakarta EE 8). 

Do we want to think ahead on how this default template would evolve with new MicroProfile+Jakarta spec levels (possibly by using new additional templates) before taking a first step here?

scottkurz:
### Changes for the v0.2.0 stack

1. Switch to UBI 8 base image.  (Fixes #259).
2. Adds Jakarta EE 8 aggregate API to default template pom.xml (Fixes #611, fixes #639).  This slows down the `appsody run` startup time a few seconds for a completely unpopulated host .m2 repo, but the difference fades once .m2 is populated.
3. Invokes the "pre-integration-test" phase before dev mode and the test goals, not just "package"
4. Adds new profiles to parent POM for each of `appsody build` (appsody-build) and host mode (default-profile), configuring the kernel artifact instead of the runtime artifact. 
5. Use "wlp.install.dir" as the Maven property, not "wlp.user.dir" to abstract over local dev mode and `appsody build`.
6. Switch to use 'default-cli' as failsafe execution.
7. Run as non-root 'java_user' in both stack image and `appsody build`.  Fixes #607.
8. Add support to no-op run/debug/test and exit '0' if a special marker file **.appsody-nodev** is detected.
9. Change `appsody run` to disable launching the OL server in debugger mode with port connected.
10. Aligns OL version in parent POM with the version tag in **Dockerfile** (Fixes #625 ), at the 19.0.0.12 fix pack.
11. Remove static/fixed, default passwords from default template for each of quickStartSecurity and keystore config and use the single generated keystore password value in both places.   Delete the quickStartSecurity config dropin, however, from the `appsody build` image in order to produce a more locked-down image.


incubator/java-openliberty/image/project/pom.xml
```diff
@@ -155,7 +155,71 @@
                             <artifactId>liberty-maven-plugin</artifactId>
                             <version>${version.liberty-maven-plugin}</version>
                             <configuration>
-                                <installDirectory>${openliberty.install.dir}/wlp</installDirectory>
+                                <installDirectory>/opt/ol/wlp</installDirectory>
```
awisniew90: Should we be using the wlp.install.dir property defined above?

incubator/java-openliberty/image/project/pom.xml
```diff
@@ -54,71 +53,66 @@
...
+            property is configured in the Liberty Maven plugin to specify 
+            where to install the Open Liberty runtime in the Docker image. --> 
         <profile>
             <id>stack-image-package</id>
             <activation>
                 <activeByDefault>false</activeByDefault>
             </activation>
             <properties>
-                <wlp.user.dir>${openliberty.install.dir}/wlp/usr</wlp.user.dir>
+                <wlp.install.dir>/opt/ol/wlp</wlp.install.dir>
```
awisniew90: Is this property used anywhere?

incubator/java-openliberty/image/project/pom.xml
```diff
@@ -155,7 +155,71 @@
...
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
+                            <groupId>io.openliberty.tools</groupId>
+                            <artifactId>liberty-maven-plugin</artifactId>
+                            <version>${version.liberty-maven-plugin}</version>
+                            <configuration>
+                                <runtimeArtifact>
+                                    <groupId>${liberty.groupId}</groupId>
+                                    <artifactId>openliberty-kernel</artifactId>
```
awisniew90: Is there a reason why we are using the kernel runtime for hostmode/build and the runtime for stack package / dev?

incubator/java-openliberty/image/project/pom.xml
scottkurz: This threw me...  the most reuse would be gained by defining a prop pointing to `/opt/ol` I was hesitant to coin a new name... so I just left it expanded.  

Besides configuring the install, the "wlp.install.dir" is used by the user to copy the project dependencies, for test, and appsody build.

incubator/java-openliberty/image/project/pom.xml
scottkurz: It's smaller, more fine-grained.  We use the bigger one in stack package/dev since we can bake it into the stack image.

Though in `appsody build`, it'd also typically be layered.. it still is a big difference to download initially (150 vs 50 MB, IIRC), and it seems now and then we've hung up Maven with the bigger zip.

Issue #631, opened 2020-02-01
Stop showing maven transfer progress for java-openliberty stack


nastacio:
**Is your feature request related to a problem? Please describe.**
The download progress for all Maven dependencies drowns out most of the useful and meaningful messages. If anything key it off from the appsody verbose tag, but don't make it the default.

**Describe the solution you'd like**
Use `-ntp` on all Maven invocations. It will cause maven to not show all the progress, but still show the summary of all downloads when complete.

**Describe alternatives you've considered**
None.

**Additional context**
None.

cc @scottkurz

scottkurz:
I think the more we can download ahead of time, the more meaningful what's left becomes, so I think this is debatable.  But, certainly a valid point to raise.

nastacio:
Maven will display eventual download problems, so there is nothing lost in not showing the progress. In fact, after writing this issue, I don't think even the `-ntp` flag shows anything all that useful.

Is it possible to tie that kind of verbose output to the usage of the `-v|--verbose` flags when invoking `appsody` ?

scottkurz:
I also find some value in knowing when I'm downloading new dependencies and hadn't thought of the verbosity as an issue.

OTOH, there is a lot of noise introduced by the /mvn/repository mount ... (I realize that's a completely different issue... but just comes to mind as we're thinking of mvn noise).

PR #655, opened 2020-02-11
Update java-microprofile stack to run as non-root

groeges:

### Checklist:

Update java-microprofile stack to run as non-root && reduce progress notification on mvn install

### Contributing a new stack:

### DIFF
--- a/incubator/java-microprofile/image/Dockerfile-stack
+++ b/incubator/java-microprofile/image/Dockerfile-stack
@@ -7,38 +7,53 @@ RUN  apt-get -qq update \
   && apt-get -qq clean \
   && rm -rf /tmp/* /var/lib/apt/lists/*
 
+RUN groupadd java_group \
+ && useradd --gid java_group --shell /bin/bash --create-home java_user \
+ && mkdir -p /mvn/repository \
+ && chown -R java_user:java_group /mvn
+
+# Dependency install
 COPY ./LICENSE /licenses/
-COPY ./project /project
-COPY ./config /config
+COPY --chown=java_user:java_group ./project /project
+COPY --chown=java_user:java_group ./config /config
+
+USER java_user
 
 RUN  /project/util/check_version build
 
 WORKDIR /project/
-RUN mkdir -p /mvn/repository
-RUN mvn -B -Dmaven.repo.local=/mvn/repository -N io.takari:maven:wrapper -Dmaven=$(mvn help:evaluate -Dexpression=maven.version -q -DforceStdout)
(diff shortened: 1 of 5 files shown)


scottkurz:
So there's three parts here:

1. Use of java_user in Dockerfile
2. Use of **-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn**
3. Changes to incubator/java-microprofile/image/Dockerfile-stack
* mounts
* java_user
* ENV APPSODY_USER_RUN_AS_LOCAL=true

For 1., I'm curious of the motive here?  Is this just a general best practice of not running as root or something more specific.

For 2., no objection but the comment mentions the Maven version in the base image not supporting:  ` --no-transfer-progress`.  I see that was added in 3.6.1 and I thought the Ubuntu is past that level.  Maybe this was added from another route without realizing this? 

For. 3, I'd like to know the goal here too.   In my experience the performance of installing Liberty into the bind-mounted **target/liberty/wlp** will be very bad..a few minutes of just sitting there while Liberty gets unzipped, and the perf. of this stack is already not good.   But the mounting is intertwined with the other changes... I'm not sure you can separate this from the other couple changes.

---

Finally, thinking out loud to myself...for the java-openliberty stack, I wonder if we need to do 1?    I don't think there's a rush for 2., though it was requested in #631.    But 3., the changes to use **java_user** is already in the works.

groeges:
Scott, let me try and answer you questions 
1) I think we are trying to modify all the stacks to run as non-root and for this reason we are creating a known user as part of the stack image creation. We don’t need to use a specific uid and gid but just create the id. Then we can setup the directories and files to be accessible by that id. One reason this was done for Kabanero is that running as non-root was a requirement for RedHat certification.

2) I tried using —no-transfer-progress and it failed. When I loaded the base image there is no maven installed hence why it then gets installled by the apt tool. In doing this it installs version 3.6.0. For this reason I used the other option to limit the output. If there is an easier way to get a later version of maven installed then happy to do that and switch to use the shorter option.

neeraj-laad:
@groeges now that we are using APPSODY_RUN_AS_LOCAL=true, we'll be running with uid/gid of the host user... Do we still need to have the java_user/java_group here? I think we still need it in the Dockerfile, but perhaps not needed in Dockerfile-build now? 

On the `--no-transfer-progress` seems much quieter and easier to read/understand so I would like to go for that and see how we get the maven upgraded. 

The change to mounts was primarily done to allow writing to `/target` from within the container. We could possibly use APPSODY_DEPS to cache that folder, but I am not sure how much change we want with his stack.

PR #661, opened 2020-02-11
Change git clone on Travis to extract labels

EnriqueL8:


.travis.yml
```diff
@@ -18,6 +18,11 @@ script:
 
 # note before_deploy will run before each deploy provider
 before_deploy:
+  - rm -rf $TRAVIS_REPO_SLUG
```
neeraj-laad: I don't believe this is the right place for this change. This will be executed after the script phase. I think we need to do this before the `script`

groeges:
@EnriqueL8 Does this mean we do 2 git clones as part of this change when we do a release build?

EnriqueL8:
Yes, because the script is in the github project... So we have to clone the project, and run the script. I was thinking we could delete the original clone after the script runs, but it's not necessary.

PR #679, opened 2020-02-20
nodejs - Optimize copy for buildah

neeraj-laad:
We are seeing significant performance issues with all Node.js based stacks to build the final application image when using buildah (instead of docker).

This is because of a known issue with buildah 1.9.0. We can not upgrade the buildah version because of other issues. 

This is a temporary fix to optimize the number of files we copy during the build process.

### Checklist:

### DIFF
--- a/incubator/nodejs/image/project/Dockerfile
+++ b/incubator/nodejs/image/project/Dockerfile
@@ -8,8 +8,8 @@ RUN apt-get update \
  && echo 'Finished installing dependencies'
 
 # Copying individual files/folders as buildah 1.9.0 does not honour .dockerignore
-COPY --chown=node:node package*.json /project/
-COPY --chown=node:node user-app /project/user-app
+COPY package*.json /project/
+COPY user-app /project/user-app
 # Removing node_modules as we can not make assumptions about file structure of user-app
  RUN rm -rf /project/user-app/node_modules && mkdir -p /project/user-app/node_modules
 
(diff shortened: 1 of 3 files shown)


sam-github:
This PR is a duplicate of https://github.com/appsody/stacks/pull/680 which changes both stacks in the same PR. I suggest closing this PR, since the changes are cut-n-paste, its easier to fix in one PR and easier to understand the historical record as well.

neeraj-laad:
@sam-github I have re-worked this PR, as we were copying `node_modules` from user-app over in the final stage of the build. This meant that dev/test dependencies from the user's development environment were getting into the final application image. 

I have now moved the copying of the application over in the first stage and I delete `node_modules` before running `npm i --production`. This has also cleaned up the copy steps a bit. 

Please let me know your thoughts.

incubator/nodejs/image/project/Dockerfile
```diff
@@ -7,11 +7,18 @@ RUN apt-get update \
  && apt-get clean \
  && echo 'Finished installing dependencies'
 
+# Copying individual files/folders as buildah 1.9.0 does not honour .dockerignore
+COPY --chown=node:node package*.json /project/
```
sam-github: These package files don't seem to be used (there is no `WORKDIR /project; RUN npm install`). The old code used to do `COPY --chown=node:node . /project`, so it would be copying `COPY package.* node_modules /project`, but that doesn't seem to be replaced. Perhaps that happens elsewhere? Sorry, I only have a general sense of how appsody uses this file.

neeraj-laad: The intent is to copy the contents of /project and /project/user-app but not copy any `/node_modules`. 

Given /project contains files from the stack, I can achieve this by copying selective files. User's appl could be in any structure so I'm copying the entire user-app folder and then delete the user-app/node_modules folder later.

That said I agree that we are not using `/project/package.json` at all. It is only used during development and adds appmetrics-dash. So we can skip copying it here.

neeraj-laad:
@sam-github Would like to get a final review from you on this.

PR #680, opened 2020-02-20
Nodejs express - optimize copy for buildah

neeraj-laad:
### Checklist:

We are seeing significant performance issues with all Node.js based stacks to build the final application image when using buildah (instead of docker).

This is because of a known issue with buildah 1.9.0. We can not upgrade the buildah version because of other issues.

This is a temporary fix to optimize the number of files we copy during the build process.

### DIFF
--- a/incubator/nodejs-express/image/project/Dockerfile
+++ b/incubator/nodejs-express/image/project/Dockerfile
@@ -8,6 +8,7 @@ RUN apt-get update \
  && echo 'Finished installing dependencies'
 
 # Install stack dependencies
+WORKDIR /project/node_modules
 WORKDIR /project
 COPY ./package*.json ./
 RUN npm install --production && tar czf stack_modules.tgz node_modules && rm -rf node_modules
--- a/incubator/nodejs-express/image/project/Dockerfile
+++ b/incubator/nodejs-express/image/project/Dockerfile
@@ -10,13 +10,13 @@ RUN apt-get update \
 # Install stack dependencies
 WORKDIR /project
 COPY ./package*.json ./
-RUN npm install --production
+RUN npm install --production && tar czf stack_modules.tgz node_modules && rm -rf node_modules
 
 # Install user-app dependencies
+RUN mkdir -p /project/user-app/node_modules
 WORKDIR /project/user-app
 COPY ./user-app/package*.json ./
-RUN npm install --production
-WORKDIR /project/user-app/node_modules
+RUN npm install --production && tar czf app_modules.tgz node_modules && rm -rf node_modules
 
 # Copy the dependencies into a slim Node docker image
 FROM node:12-slim
@@ -28,13 +28,19 @@ RUN apt-get update \
  && echo 'Finished installing dependencies'
 
 # Copy the project
-COPY --chown=node:node . /project
+COPY --chown=node:node package*.json /project/
+COPY --chown=node:node *.js /project/
+COPY --chown=node:node user-app /project/user-app
 
-# Copy all dependencies
-COPY --chown=node:node --from=0 /project/node_modules /project/node_modules
-COPY --chown=node:node --from=0 /project/user-app/node_modules /project/user-app/node_modules
+# Copy all app dependencies
+WORKDIR /project/user-app
+COPY --chown=node:node --from=0 /project/user-app/app_modules.tgz .
+RUN tar xf app_modules.tgz && chown -R node:node node_modules && rm app_modules.tgz
 
+# Copy all stack dependencies
 WORKDIR /project
+COPY --chown=node:node --from=0 /project/stack_modules.tgz .
+RUN tar xf stack_modules.tgz && chown -R node:node node_modules && rm stack_modules.tgz
 
 ENV NODE_PATH=/project/user-app/node_modules
 
--- a/incubator/nodejs-express/image/project/Dockerfile
+++ b/incubator/nodejs-express/image/project/Dockerfile
@@ -7,21 +7,23 @@ RUN apt-get update \
  && apt-get clean \
  && echo 'Finished installing dependencies'
 
+# Copying individual files/folders as buildah 1.9.0 does not honour .dockerignore
+COPY package*.json ./
+COPY *.js /project/
+COPY user-app /project/user-app
+# Removing node_modules as we can not make assumptions about file structure of user-app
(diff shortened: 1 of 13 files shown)


incubator/nodejs-express/image/project/Dockerfile
```diff
@@ -28,13 +29,19 @@ RUN apt-get update \
  && echo 'Finished installing dependencies'
 
 # Copy the project
-COPY --chown=node:node . /project
```
sam-github: Does buildah do better  when copying between images than it does when copying from host?

It looks like all the lines from here to `ENV NODE_PATH` could be replaced by a single line:

> COPY --chown=node:node --from=0 /project /project

EnriqueL8: In the first docker image we don't have all the files from the `/project` dir, since we only copy the `package.json` for the project and the user-app. I agree that we could simplify the process by first copying the tars, and `package.json'`s with that one command and then copy the files from the host.

incubator/nodejs/image/project/Dockerfile
```diff
@@ -7,11 +7,12 @@ RUN apt-get update \
  && apt-get clean \
  && echo 'Finished installing dependencies'
 
+WORKDIR "/project/user-app/node_modules"
```
EnriqueL8: Are the `nodejs` specific changes in the other PR also?

incubator/nodejs/stack.yaml
```diff
@@ -1,5 +1,5 @@
 name: Node.js
```
EnriqueL8: Also this change, should be remove since we have another PR for nodejs

incubator/nodejs/image/project/Dockerfile
sam-github: Yes: https://github.com/appsody/stacks/pull/679#issuecomment-589350889

incubator/nodejs-express/image/project/Dockerfile
```diff
@@ -8,15 +8,16 @@ RUN apt-get update \
  && echo 'Finished installing dependencies'
 
 # Install stack dependencies
+WORKDIR /project/node_modules
```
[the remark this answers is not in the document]
neeraj-laad: The intent was to always create the node_modules folder as if there are no dependencies then it might not get created and then the tar commands could fail. But in this case, this is not really an issue as we know exactly what is in the stack. 

Deleting this line.

incubator/nodejs-express/image/project/Dockerfile
```diff
@@ -8,15 +8,16 @@ RUN apt-get update \
...
 WORKDIR /project
 COPY ./package*.json ./
-RUN npm install --production
+RUN npm install --production && tar czf stack_modules.tgz node_modules && rm -rf node_modules
 
 # Install user-app dependencies
+WORKDIR /project/user-app/node_modules
```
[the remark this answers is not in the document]
neeraj-laad: Wil change to `RUN mkdir -p <path>` to be clear about what we are trying to do here.

incubator/nodejs-express/image/project/Dockerfile
neeraj-laad: The issue we have today is that buildah 1.9.0 does not honour `.dockerignore`, and has a buggy algorithm for iterating over files. We can not move to a higher version (which has this fix) yet because of other issues.

I can move the copy commands in the first stage, however, we will still need those separate copy statements in the first stage then. Would we just move the problem around to the first stage?

neeraj-laad:
@sam-github Thanks for the review comments. I accidentally created the PR with changes to multiple stacks. It is easier to release stacks individually so have split the changes in 2 PRs again.

I have added comments to provide the link to the PR that has further links to original issue. I hope that helps provide some context to the issue.

sam-github:
> The issue we have today is that buildah 1.9.0 does not honour .dockerignore, and has a buggy algorithm for iterating over files. We can not move to a higher version (which has this fix) yet because of other issues.

^--- without this information, the Dockerfile isn't understandable on its own, so its likely to get refactored to break buildah. I strongly suggest that the information above be put into the dockerfile in a comment. It doesn't necessarily need a link to this PR in the comment, since that can be found from `git blame`.

Also, I'm a bit confused by "we can't update", who is "we"? appsody doesn't use buildah, does it? Since buildah isn't used in appsody AFAICT, its not possible to know if any future change here breaks whoever is using buildah downstream (who is using buildah downstream?). At some point in the future, I assume this structure can be deleted, but without any reproduction information here, or link to the downstream consumer who is having this issue, it won't be possible to know or ask when the workaround can be deleted.

neeraj-laad:
Appsody uses buildah instead of docker when used in a Kubernetes environment like Che hosted IDE and Tekton build pipelines. appsody/tekton-sample repo even provides sample Tekton pipelines for user's to start with.

The version of buildah that we use today 1.9.0 has this performance issue during copy. Newer version of buildah have resolved this but they have only works if newer versions of  `fuseoverlayfs` is installed on the host (not the container). Given the changes are needed on the host it makes it difficult for Apprody team to control and fix.

neeraj-laad:
@sam-github I have re-worked this PR as we were copying node_moduels from user's app over in the final stage. This meant that dev/test dependencies fro muser's app were still getting into the final application image. 

I have now moved the copying of the application over in the first stage and I delete node_modules before running `npm i --production`. 

This has also cleaned up the copy steps a bit. 

Please let me know your thoughts.

incubator/nodejs-express/image/project/Dockerfile
```diff
@@ -7,16 +7,23 @@ RUN apt-get update \
  && apt-get clean \
  && echo 'Finished installing dependencies'
 
+# Copying individual files/folders as buildah 1.9.0 does not honour .dockerignore
+COPY package*.json ./
```
groeges: @neeraj-laad Is this COPY correct? Where are the package*.json files getting copied to?
Where this was moved from we had done a `WORKDIR /project` first so I presume the .json files were being copied into /project, but with it as we have now are they getting copied into the correct directory.

neeraj-laad: Good catch..

neeraj-laad:
@sam-github Would like to get a final review from you on this.

PR #684, opened 2020-02-20
openliberty - Add ARG to stack Dockerfiles to allow appsody to pass docker-options for build-arg JAVA_OPTIONS

raguks:
### Checklist:

We are adding the IBM_JAVA_OPTION to dockerfile so when appsody --docker-options pass --build-args to docker build for apps under development, JAVA_OPTIONS for OpenJ9 jvm are available in the liberty app container that runs on openshift/k8s. This not only helps our appsody goals but all real enterprise application configuration. This allows customers to pass all kinds of options they would need to set to be able to run a real enterprise app. The actual passing of args via appsody is still difficult given all the escape sequences that need to be used. 

### Related Issues:

https://github.com/appsody/appsody/issues/900

### DIFF
--- a/incubator/java-openliberty/image/project/Dockerfile
+++ b/incubator/java-openliberty/image/project/Dockerfile
@@ -60,6 +60,9 @@ RUN cd /project/user-app/target/liberty/wlp/usr/servers && \
 # Step 2: Package Open Liberty image
 FROM openliberty/open-liberty:{{.stack.libertyversion}}-kernel-java8-openj9-ubi
 
+# Allow appsody deploy to set --build-arg for an env variable
+ARG IBM_JAVA_OPTIONS
+
 #2a) copy any resources 
 COPY --chown=1001:0 --from=compile /shared /opt/ol/wlp/usr/shared/
 
--- a/incubator/java-openliberty/stack.yaml
+++ b/incubator/java-openliberty/stack.yaml
@@ -1,5 +1,5 @@
 name: Open Liberty
-version: 0.2.1
+version: 0.2.2
 description: Open Liberty & OpenJ9 using Maven
 license: Apache-2.0
 language: java


raguks:
Please review and merge. Thank you.

raguks:
@EnriqueL8 @scottkurz can you please review and merge.

scottkurz:
Thx @raguks for the PR>

I'm curious if you could give a bit more detail about what kind of usages you'd expect here?

Is it simply general.. to just want to expose the options variable in its full, general form (with all the options)?    I'm just wondering if you have something more specific in mind, like say, 80% of the time they're going to be setting min/max heap size (assuming that's even one of the options).

I'd think that, for production, some of the values here would not be applied at build time but only when running the image, via env var or ConfigMap.   So again, just trying to build up some idea of the use cases for which you'd want to build these options into the image.    (Note we already use some of the options related to shared class caching).

(Note also the value is being deprecated and replaced with the **OPENJ9_JAVA_OPTIONS** var, but we're not quite at that level yet.) 

Thanks

raguks:
Thank you @scottkurz for letting me know about OPENJ9_JAVA_OPTIONS - I was hoping we will get there at some point! And on what we are trying to achieve:

We have a way of running openj9 where JIT compilation can be done outside jvm. This is a new feature you might already know. For openliberty and microprofile applications to enable their clients to use remote JIT compiler (JIT Server) we need to update the JAVA_OPTIONS. Exact commands are in appsody/appsody#900. Because this works like a switch and we need to do appsody/kabenero integration, we need a way to do it at deploy time (Ours is a way of deployment of apps).

Also, I was thinking in general that appsody and stacks should have a way to set Deployment options at deploy time rather than as part of the development config. Something you may have already thought of. I am working with appsody teams to get some bugs fixed to be able to do this. They have been helpful and fixed a bug - one more pending to make this deploy time config work at appsody level - then these two PRs(appsody/stacks#684
appsody/stacks#685) will help make it work at stack level. Please let me know if you have any questions.

scottkurz:
@raguks that sounds interesting...

I just question if we should update the stack with a build ARG or give `appsody deploy` another means to provide the env var at runtime.    

Made a comment here: https://github.com/appsody/appsody/issues/900#issuecomment-591523745

raguks:
@scottkurz 
1. I thought you are the stack author - if not can you please check with a stack author?
2. Can you suggest some ways to pass env variables to a container without adding ARG to Dockerfile? I think it is a Docker requirement. It would be happy to use it if there is an alternative.
3. Can you elaborate on JAVA_OPTIONS being a build variable? As long as env are appended and not overwritten we should be fine.

scottkurz:
@raguks 
>     1. I thought you are the stack author - if not can you please check with a stack author?

I am one of the stack authors.
 
>     2. Can you suggest some ways to pass env variables to a container without adding ARG to Dockerfile? I think it is a Docker requirement. It would be happy to use it if there is an alternative.
>     3. Can you elaborate on JAVA_OPTIONS being a build variable? As long as env are appended and not overwritten we should be fine.

@neeraj-laad  can you please took a look at this?   As I mentioned in https://github.com/appsody/appsody/issues/900,  the IBM_JAVA_OPTIONS env var is useful at both build time and run time (and in fact I think the runtime value gets appended to the build time value).  

So I recognize exposing IBM_JAVA_OPTIONS as a build arg would allow someone to set the values they need on `appsody deploy`.   My point is that, say you had the need to set an env var that was only used at runtime and not at build time... do we have a way to do that in `appsody deploy`?

If so maybe @raguks can use that.  If not, maybe we might want to add that to `appsody deploy` anyway.

neeraj-laad:
@scottkurz I added a comment here https://github.com/appsody/appsody/issues/900#issuecomment-592468385. Please let me know if I have not answered your question.

I think we have a solution for deployment stage already, but we don't really have a way to do this at application build time and stack creator has to explicitly allow/disallow what developer can influence.

scottkurz:
Thx @neeraj-laad  you have answered my question !  

I'm noticing that we do not expose IBM_JAVA_OPTIONS as a build arg in the base Open Liberty images that we build from in this stack, (with the source Dockerfile(s) coming from [this](https://github.com/OpenLiberty/ci.docker) project).   

So following this as a pattern, I think we could argue that supplying the values at `appsody run`  and `appsody deploy` would be enough.

So @raguks, am I convincing you?    If you think there's a strong case for supplying this value at image build time then you should probably open an issue there, too, for non-appsody cases.

raguks:
@scottkurz can you please get the stack owners to present a use case on how open-liberty can be used with and without JAVA OPTIONS in a K8 environment? These env have to be updated by developers before deploying. So if we include then in Dockerfile or not is relevant if we know developers dont add it later. Who owns the liberty stack can we please bring them in for the discussion here?

scottkurz:
@raguks, I am the stack owner.    

From Neeraj's comment: https://github.com/appsody/appsody/issues/900#issuecomment-592468385

I'm suggesting you could:

> edit the deployment manifest and provide environment variables etc in the manifest before running the deploy.

instead of building them into the image. 

Do you see this working for you?

raguks:
@scottkurz This should work if the env set before liberty starts. I am trying this.

scottkurz:
For now I'm going to close this and prefer the approach of updating the app-deploy.yaml rather than supporting this at build time.

Issue #692, opened 2020-02-24
nodejs: add a stack for Kafka


ianpartridge:
**Is your feature request related to a problem? Please describe.**

Currently the `nodejs-express` stacks provides 2 templates: `simple` and `scaffold`.

We would like to add support for event-driven applications using Kafka. This should help users quickly get started with building Kafka-enabled applications, and show how to do consume/produce on Kafka topics.

This could be delivered as an additional template, or a new stack.

**Additional context**

Assess these libraries:
https://github.com/SOHU-Co/kafka-node
https://github.com/tulios/kafkajs
https://github.com/Blizzard/node-rdkafka

ianpartridge:
^^^ @sam-github :)

ianpartridge:
I spoke to @sam-github about this.  As Express is focussed on HTTP routing it probably wouldn't make sense to have a Kafka template in that stack.

The best thing may be a new `nodejs-kafka` stack, we would need to decide which NPM package to use for the Kafka connector (thish would need an API assessment).

Issue #695, opened 2020-02-26
spring: add Kafka-enabled template/stack


ianpartridge:
**Is your feature request related to a problem? Please describe.**

Currently, developers who want to build Spring applications using Appsody which talk to Kafka have to roll everything themselves.

**Describe the solution you'd like**

An easy way for developers to build event-driven Kafka app on Spring, with Appsody.

**Additional context**

This could be done through [spring-kafka](https://spring.io/projects/spring-kafka) or [spring-cloud-stream](https://spring.io/projects/spring-cloud-stream).

richard-trotter:
Here is a sample app that "illustrates the use of an IBM Event Streams Spring Boot "starter" to assist Event Streams service integration"

Using Appsody in Spring Microservice Development for Kubernetes
https://github.com/richard-trotter/appsody4springsvc2k8s

As I recall, the work required on the part of the developer is:
1. add POM dependencies
2. add configuration properties to application.yaml
3. create a K8s Secret with Event Streams credentials
4. implement the application code, integrating via org.springframework.kafka API

ianpartridge:
Nice, those are certainly things we can help developers with via a bespoke stack/template.

NB. service binding to Strimzi is planned to be via https://github.com/redhat-developer/service-binding-operator (@arthurdm is on point there) :)

ianpartridge:
This was delivered via https://github.com/appsody/stacks/pull/742 :)

PR #698, opened 2020-03-03
nodejs-express: new template

sam-github:
Fixes: https://github.com/appsody/stacks/issues/692
Related: https://github.com/appsody/stacks/issues/695

### Checklist:

- [ ] Read the [Code of Conduct](https://github.com/appsody/website/blob/master/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md).

- [ ] Followed the [commit message guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md#commit-message-guidelines).

### Contributing a new stack:


sam-github:
@ianpartridge PTAL

incubator/nodejs-kafka/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Application metrics
+
+Enable powerful monitoring for your distributed application and configure rule-based alerting using Prometheus. This is vital for diagnosing problems and ensuring the reliability of your application.
+
+The [appmetrics-prometheus](https://github.com/CloudNativeJS/appmetrics-prometheus) module will collect a wide range of resource-centric (CPU, memory) and application-centric (HTTP request responsiveness) metrics from your application, and then expose them as multi-dimensional time-series data through an application endpoint for Prometheus to scrape and aggregate.
+
+This stack also comes with Prometheus metrics, which has been preconfigured to work with your application. You will not be able to override this endpoint:
+
+- Metrics endpoint: http://localhost:3000/metrics
```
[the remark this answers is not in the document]
sam-github: No, sorry. Once we move from appmetrics to prom-client, then user-defined metrics will be possible, and the user will be able to add whatever metrics they find meaningful.

incubator/nodejs-kafka/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The application must have a top-level package.json file, and it must contain a
+property `"main": APP` (the templates all set `APP` to `"app.js"` and include a
+skeleton implementation in that file). The `APP` will be required, and must
+export a function.
+
+The function will be called with an `options` argument, which may be ignored.
+
+Options contains:
+- `Kafka` {node-rdkafka} Use this to configure and use node-rdkafka.
```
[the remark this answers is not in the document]
sam-github: Intentionally, its `Kafka` in all the node-rdkafka examples and API documentation, so I kept it upper-case here to emphasize its non-magical nature - it is identical to what they would get if they required it, but the dependency is injected (so version controlled by the stack). That said said, I've no strong feelings, I could lower-case it for internal consistency with the other injected properties.

incubator/nodejs-express/image/project/Dockerfile
```diff
@@ -47,4 +47,5 @@ ENV PORT 3000
 USER node
 
 EXPOSE 3000
+# Where is WORKDIR or CWD set?
```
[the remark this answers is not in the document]
sam-github: Mistake. I'll remove, it was a note to look into a bug that got fixed in nodejs-express and since stacks don't share content easily, I had to clone the fix into the Kafka stack after rebasing.

incubator/nodejs-kafka/templates/default/app.js
```diff
@@ -0,0 +1,115 @@
...
+  // options.log is a pino logger:
+  // - https://github.com/pinojs/pino/blob/master/docs/api.md#logger
+  const log = options.log;
+
+  // options.app is an express instance, routes can be configured.
+  const app = options.app;
+
+  // KAFKA_BOOTSTRAP_SERVERS can be injected by the Service Binding Operator
+  // - https://github.com/redhat-developer/service-binding-operator
+  const brokers = process.env.KAFKA_BOOTSTRAP_SERVERS || 'localhost:9092'
```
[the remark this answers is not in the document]
sam-github: I can bake it in and expose it as `options.brokers`. It seemed like the name might be an ephemeral detail of WIP examples.  "Servers" are called "brokers" by Kafka, and I've no idea what "bootstrap" means in this context. Could we call it `KAFKA_BROKERS` (Kafka's term) or `KAFKA_SERVICES` (k8s' term)? If not, `KAFKA_BOOTSTRAP_SERVERS` it can be.

incubator/nodejs-kafka/README.md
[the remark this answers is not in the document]
ianpartridge: Sad times. OK :)

incubator/nodejs-kafka/README.md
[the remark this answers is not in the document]
ianpartridge: OK cool, I'm not a Node person so whatever Node developers will feel happy with :)

incubator/nodejs-kafka/templates/default/app.js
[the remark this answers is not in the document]
ianpartridge: Yeah I think this is a short-term stop gap while the details of how Service Binding works are hammered out.

For example we'll need to support the binding information that Event Streams uses, which is not just host:port but also TLS-enabled and with authentication via SASL-SCRAM.

One simple env var is OK now (any maybe we do leave it in the template) and we can revisitit in the future.

incubator/nodejs-kafka/README.md
```diff
@@ -0,0 +1,88 @@
+# Node.js Kafka Stack
+
+The Node.js Kafka stack extends the [Node.js stack](../nodejs/README.md) and provides a consistent way of developing applications using [node-rdkafka](https://www.npmjs.com/package/node-rdkafka) and [Express](http://expressjs.com).
+
+This stack is based on `Node.js v12` and `node-rdkafka v2.7.x` and enables health checking and application metrics out of the box. The stack also provides a set of tests written using the [Mocha](https://mochajs.org) test framework, while the application metadata is provided via a `package.json` file.
```
richardlau: It's probably not an immediate issue as this is using Node.js 12, but I recently came across `node-rdkafka` as somebody raised an issue in nodejs/node-gyp (https://github.com/nodejs/node-gyp/issues/2055) as it wasn't compiling with Node.js 13. I didn't do a deep search (just enough to demonstrate it wasn't a node-gyp issue) and found https://github.com/Blizzard/node-rdkafka/issues/733.

ianpartridge: Thanks Richard. I think we're fine to use this library on Node.js 12 for now.

sam-github: I got it building on 13.x. This is still incubator, and we're doing some ongoing research into what the "best" Kafka library is a the moment. The situation is fluid :-).

sam-github:
ping @neeraj-laad : https://github.com/appsody/stacks/pull/698#issuecomment-595485498

sam-github:
Strange, node-rdkafka failed to build after master was merged into this branch. I'll look into it.

neeraj-laad:
@sam-github just curious to understand why is this stack inheriting from `nodejs` and not `nodejs-express` stack as it has node express and kafka?

neeraj-laad:
Regarding your earlier comment on experimental vs incubator. Appsody does not mandate that all stacks start as experimental. As long as you are willing to maintain and try to get it to stable. However, if we are not 100% sure if this is a good idea or taking shortcuts for some behaviour, then we can put this in experimental too.

sam-github:
> just curious to understand why is this stack inheriting from nodejs and not nodejs-express stack as it has node express and kafka?

I'll see if I can slim the stack down via inheritance, but I'm not sure it matters. There don't look to be any layers that are fruitfully inherited. Dockerfiles still have to contain the same contents, the /package gets overwritten and has different contents, the templates are different, the READMEs are self-contained. If there was some way of making the lower-stack layers contain things that inherited stacks could just use, inheritance would be more meaningful, but either because of the nature of node.js apps, or just the nature of stacks that they _have_ a top-level app contained in them that then delegates some things to the template, I don't see a lot of inheriting going on.

Even nodejs-express contains, AFAICT, basically just a verbatim  copy of much of the contents of nodejs, even though it inherits from it.

But maybe the stacks weren't written with enough attention towards inheritance. I'll see.

sam-github:
I moved it to experimental because we don't know if we can commit to node-librdkafka for the longer term, even though it is best choice ATM. 

I copied the recent changes from the other nodejs stacks to not update system or npm packages.

I'd prefer to leave the inheritance change, if it ends up being useful, to happen as part of a greater cleanup I'm working on.

sam-github:
OK, passing, PTAL @neeraj-laad

experimental/nodejs-kafka/README.md
```diff
@@ -0,0 +1,88 @@
...
+1. You should be able to access the following endpoints, as they are exposed by your template application by default:
+
+    - Health endpoint: http://localhost:3000/health
+    - Liveness endpoint: http://localhost:3000/live
+    - Readiness endpoint: http://localhost:3000/ready
+    - Metrics endpoint: http://localhost:3000/metrics
+
+1.  A message may be posted into the default skeleton app. This could be done with `curl`:
+    ```bash
+    curl -v -X POST -H "content-type: application/json" \
```
Kamran64: Running this command at first succeeds for me although eventually it looks like connection to the producer/consumer is lost:
```
[Container] {"level":50,"time":1584971907527,"pid":45,"hostname":"f4395b2e3ac3","msg":"error from consumer: Error: broker transport failure","v":1}
[Container] {"level":50,"time":1584971907527,"pid":45,"hostname":"f4395b2e3ac3","msg":"error from consumer: Error: all broker connections are down","v":1}
```
which then results in failures running this. 

Also nit: You could cut out the use of the -X flag. Not a big deal but I get this note when running the command:
```
Note: Unnecessary use of -X or --request, POST is already inferred.
```

ianpartridge: I see this problem too.  Are you running the https://github.com/lensesio/fast-data-dev image that Sam suggested?  I also tried using a `docker-compose.yaml` for Kafka that I had lying around, but that has the same problem.

ianpartridge:
Can I suggest we change the README away from the Lenses.io image, and use Docker Compose running Strimzi instead - https://github.com/scholzj/strimzi-compose-up

```yaml
version: '2'

services:

  zookeeper:
    image: strimzi/kafka:0.15.0-kafka-2.3.1
    command: [
      "sh", "-c",
      "bin/zookeeper-server-start.sh config/zookeeper.properties"
    ]
    ports:
      - "2181:2181"
    environment:
      LOG_DIR: /tmp/logs

  kafka:
    image: strimzi/kafka:0.15.0-kafka-2.3.1
    command: [
      "sh", "-c",
      "bin/kafka-server-start.sh config/server.properties --override listeners=$${KAFKA_LISTENERS} --override advertised.listeners=$${KAFKA_ADVERTISED_LISTENERS} --override zookeeper.connect=$${KAFKA_ZOOKEEPER_CONNECT}"
    ]
    depends_on:
      - zookeeper
    ports:
      - "9092:9092"
    environment:
      LOG_DIR: "/tmp/logs"
      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092
      KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
```

sam-github:
READMEs don't usually describe how to setup the backend services, I think I should remove the note, anyone contemplating using Kafka should already have their favourite way to run it.

sam-github:
I'm still looking at how to fix this, but at this point I've convinced myself its a network config issue, I removed the appsody layer and just ran the scaffolded app.js locally on my host, and it works fine, but doesn't inside appsody run's docker container.

cf. https://www.confluent.io/blog/kafka-listeners-explained/, the 9092 port is reachable, the client connects, then drops the connection. I think we hit a variant of "Why can I connect to the broker, but the client still fails?".

sam-github:
@Kamran64 So, there are two issues here, the first of which seems a generic problem, how does someone usually run a container in development with `appsody run` and have it reach a dependent service that is also running in docker-for-desktop? This is kafka, but mysql, etc, would have the same issue AFAICT.

- the `appsody run` container can't reach the containers started by docker-compose, they aren't on the same network, and `appsody run --docker-options` explicitly documents `--network` as not supported

On Linux, this works, because `--net=host` runs the container on the host network, but on docker-for-desktop it either does nothing, or runs it on the *VM*'s "host" network, but that doesn't help the mac.

- once run, the container needs to know its *external* name, and advertise it, this just needs adding to the kafka env.

The general recommendation here seems to be "use docker-compose", which I did, and it works, but not with `appsody run`. See: https://github.com/sam-github/appsody-stacks/blob/a42823654767141a7e4b6bcd8279fa3fa5d09454/experimental/nodejs-kafka/templates/default/docker-compose.yaml#L16, which assumes you've done a `mkdir ex; cd ex; appsody init dev.local/nodejs-kafka; appsody build` to create the image that will be run by docker-compose. You can curl it from the (mac) host.


The kube approach would be to deploy kafka and zookeeper into kube, create a service for kafka, and use that service name to connect from the appsody nodejs-kafka container (in production) and in dev, do the same, except do a `kubectl port-forward` to expose the kafka service onto the local host.

sam-github:
@neeraj-laad ^--- see above, what is appsody recommendation for a container run with `appsody run` on OS X docker-for-desktop to be able to reach a service in another container in docker-for-desktop? General recommendation is to "use docker-compose for local dev", but appsody is going a different direction.

experimental/nodejs-kafka/templates/default/app.js
```diff
@@ -0,0 +1,143 @@
...
+
+  log.info('Kafka features: %s', Kafka.features);
+  log.info('Kafka brokers: %s', brokers);
+
+  // Based on:
+  // - https://github.com/ibm-messaging/event-streams-samples/blob/master/kafka-nodejs-console-sample/app.js#L145
+  const config = {
+    // debug: 'all',  // Setting will cause 'event.log' events.
+    'metadata.broker.list': brokers,
+    // 'log.connection.close' : false, // Why?
```
sam-github, line 28: @dalelane the event-streams-sample does this, but its not clear why, even after reading the docs at https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

experimental/nodejs-kafka/templates/default/app.js
```diff
@@ -0,0 +1,143 @@
...
+
+// Based on:
+// - https://github.com/Blizzard/node-rdkafka/blob/master/examples/producer.md
+// - https://github.com/ibm-messaging/event-streams-samples/blob/master/kafka-nodejs-console-sample/producerLoop.js
+function Produce(Kafka, config, log, app) {
+  config.dr_msg_cb = true;  // Enable delivery reports with message payload
+  var producer = new Kafka.Producer(config);
+  var topicName = 'my-topic';
+
+  producer.setPollInterval(100); // Why?
```
sam-github, line 43: @dalelane the example also does this, also not clear why

experimental/nodejs-kafka/templates/default/app.js
```diff
@@ -0,0 +1,114 @@
+const express = require('express');
```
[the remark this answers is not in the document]
sam-github: added a hello world

sam-github:
I'm still not a Kafka install expert, but @dalelane of the event team helps maintain https://github.com/strimzi/strimzi-kafka-operator#quickstart, perhaps we should link to it in the README?

sam-github:
https://github.com/appsody/stacks/blob/fc83e6b1722054f9b809337c57da5e41565c4341/incubator/java-spring-boot2/templates/kafka/README.md  <--- perhaps this should be aligned with the java kafka stack in its README? And possibly in what it does in the default template?

sam-github:
@ianpartridge I see https://github.com/appsody/stacks/pull/742 landed, and they figured out how to use appsody run with kafka, I synced the instructions here with that PR.

PTAL

sam-github:
Turns out its become a requirement that all the kafkas be templates, I copied the README from the java-spring-boot2 (basically), and made this a template.

The libssl deps can only be added in the stack's dockerfile, and nodejs-express stack (and likely all of the nodejs stacks) doesn't support the user's app depending on compiled addons (permissions problems on the stack user's home directory, /root), which is why rdkafka was added as a dep of the image/project (fixing this bug in user setup is out of scope for this PR).

to: @ianpartridge

incubator/nodejs-express/templates/kafka/README.md
```diff
@@ -0,0 +1,59 @@
...
+1. To deploy the application to Kubernetes run the following command:
+    ```bash
+    appsody deploy
+    ```
+    Make sure to add the `KAFKA_BOOTSTRAP_SERVERS` environment variable in the `app-deploy.yaml` before running the above command
+
+    ```yaml
+    env:
+      - name: KAFKA_BOOTSTRAP_SERVERS
+	value: ${KAFKA_BOOTSTRAP_SERVERS}
```
ianpartridge, line 44: Tab at the start of this - should be spaces.

incubator/nodejs-express/templates/kafka/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "nodejs-kafka-consumer",
```
ianpartridge, line 2: Not just consumer?

incubator/nodejs-express/templates/kafka/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "0.1.0",
+  "description": "Node.js Kafka consumer",
+  "license": "Apache-2.0",
+  "main": "app.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/appsody/stacks.git",
+    "directory": "incubator/nodejs-kafka/templates/consumer"
```
ianpartridge, line 10: `incubator/nodejs-express/templates/kafka` I think.

incubator/nodejs-express/templates/kafka/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "nodejs-kafka-consumer",
+  "version": "0.1.0",
+  "description": "Node.js Kafka consumer",
```
ianpartridge, line 4: Remove "consumer"?

sam-github:
`templates/kafka/package.json` made consistent with `templates/*/package.json`

copied in docker-compose file

PTAL @ianpartridge

sam-github:
@neeraj-laad Can you land?

neeraj-laad:
Once we bump the stack version, this is good to go

sam-github:
@Kamran64 done

PR #700, opened 2020-03-04
Update to build nodejs-express stack as non-root

groeges:

### Checklist:

### Contributing a new stack:

Updates to Dockerfiles to allow stacks to be built as non-root

### DIFF
--- a/incubator/nodejs-express/image/Dockerfile-stack
+++ b/incubator/nodejs-express/image/Dockerfile-stack
@@ -1,5 +1,7 @@
 FROM appsody/nodejs:0.3
 
+RUN useradd --gid 0 --shell /bin/bash --create-home node_user
+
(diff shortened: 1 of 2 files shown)


sam-github:
I suggest that instead, we do a `RUN usermod -g 0 node` in the nodejs stack, so that all the nodejs stacks inherit the group 0 behaviour.

For UBI images, they don't have a `node` user, they seem to have:

> default:x:1001:0:Default Application User:/opt/app-root/src:/sbin/nologin

To minimize the distance between the stack forks, is it worth trying to reconcile some of this?

Its possible the kabanero fork of the nodejs stack could create a node user (`RUN useradd -u 1000 -g 0 ... node; USER node;`) so the derived stacks are the same, or maybe nodejs should be doing as ubi does, making a default user and calling `USER default`.

PR #718, opened 2020-03-13
kitura: Update to not upgrade packages in base image

groeges:

### Checklist:

### Contributing a new stack:

### DIFF
--- a/incubator/kitura/image/project/Dockerfile
+++ b/incubator/kitura/image/project/Dockerfile
@@ -24,11 +24,6 @@ RUN echo '#!/bin/bash' > run.sh \
 
 FROM swift:5.1-slim
 
-# Install OS updates
-RUN apt-get update \
- && apt-get clean \
- && echo 'Finished installing dependencies'
-
 # Define a kitura user
 RUN useradd -ms /bin/bash kitura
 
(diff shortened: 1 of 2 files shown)


ianpartridge:
Sorry, could you explain the rationale for this? :)

groeges:
@ianpartridge We have had issue recently where the application dockerfile build failed due to updates being made in some packages that were updated since the stack image had been tested / created.  

As such, we have been removing all the places in the dockerfile where an apt-get upgrade is being done in order to keep just the package versions available in the base stack.  

Looking at the changes I may have  misread "update" as "upgrade" and remove those lines, as this PR might not be valid, but does just doing an "apt-get update && apt-get clean" actually do anything? If not, does it need to be done?

ianpartridge:
Thanks @groeges.

"`apt-get update`" just updates the list of packages and their versions from the package database, it does not actually upgrade any packages to newer versions. "`apt-get upgrade`" actually upgrades the packages. As such, the current message of "Finished installing dependencies" is incorrect and misleading.

So, I think we should merge this PR. If we ever want to explicitly upgrade packages in the future we can re-add the `apt-get update` at that time.

N.B. It is important to run `apt-get update` before `apt-get upgrade` so that you are upgrading to the latest version of the package, not whatever the latest version in the local package index was.

PR #727, opened 2020-03-23
Fix download cli

EnriqueL8:
### Checklist:

### Issue
Travis is hitting a limit of https requests to Github to download the CLI. Adding read only token to increase this limit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ go:
 os: linux
 dist: bionic
 
+env:
+  - GITHUB_READ_TOKEN: "05a26637e299a0d3481c9f9c62c8d5a373c74af9"
+
 services:
   - docker
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ go:
 os: linux
 dist: bionic
 
-env:
-  - GITHUB_READ_TOKEN: "05a26637e299a0d3481c9f9c62c8d5a373c74af9"
-
 services:
   - docker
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,6 +6,9 @@ go:
 os: linux
 dist: bionic
 
+env:
+  - GITHUB_READ_TOKEN: "05a26637e299a0d3481c9f9c62c8d5a373c74af9"
```
sam-github: > Warning: Treat your tokens like passwords and keep them secret. When working with the API, use tokens as environment variables instead of hardcoding them into your programs.
- https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line

In this case, shouldn't this be in the Travis setup so that it is not public? Though maybe it should be revoked now that it is public.

EnriqueL8: This token is a read-only token (no scope) - see [oauth-scopes](https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/.) Recently we believe github has introduced a limit to it's `HTTP`calls which means that if we run a significant number of Travis runs we can hit this limit. We can increase the number of request by adding a token, and the only way to add a token to PRs which have been made from an external fork is to expose them in the `.travis.yml`. (Travis automatically detects a token if added through the settings and doesn't expose it)

I agree, it is not a nice solution but since this token has very limited permissions I believe it's not an extreme security risk. We also store a token used in the release process, which is set in the Travis setup and hidden from users.

EnriqueL8: Update: Putting token in travis setting, but will still be visible in the travis build on PRs

neeraj-laad: This is better than having the token in travis.yaml.

PR #738, opened 2020-03-27
feat: upgrade nodejs-loopback stack to the latest version of LoopBack

raymondfeng:
This PR upgrade `nodejs-loopback` stack to use latest version of LoopBack.

### Contributing a new stack:


raymondfeng:
> Looks good but the version needs to be bumped in the package.json also

Fixed.

Issue #518, opened 2019-11-22
nodejs-express: appsody run creates a root owned node_modules in on dev host


sam-github:
**Describe the bug**

`appsody run` causes a `docker build`, which causes a `RUN npm install` as root, which creates a `node_modules` on the dev host with root ownership. That root owned directory makes it difficult for the user to do things like install their app's dependencies.

**To Reproduce**

```
cloud-node/ex-ext2 % rm -rf node_modules                                               
..... appsody run.... ctrl-C (or not)
cloud-node/ex-ext2 % ls -l                                                             
...
drwxr-xr-x  2 root root  4096 Nov 22 12:53 node_modules/                               
-rwxrwxr-x  1 sam  sam    506 Nov 22 12:49 package.json*
...
cloud-node/ex-ext2 % npm install --save lodash           
npm WARN checkPermissions Missing write access to /home/sam/w/cloud-node/ex-ext2/node_modules
npm ERR! code EACCES                                                                   
npm ERR! syscall access
npm ERR! path /home/sam/w/cloud-node/ex-ext2/node_modules                              
npm ERR! errno -13
npm ERR! Error: EACCES: permission denied, access '/home/sam/w/cloud-node/ex-ext2/node_modules'
```

**Expected behavior**

To be able add a new dependency to my project.

**Actual behaviour**
What is the actual behaviour.

It failed to do so.

**Environment Details (please complete the following information):**
 - OS: [e.g. iOS] Ubuntu 19.10
 - Version [e.g. 22]  Version of what?

**If applicable please specify:**
 - CLI version: 0.5.0
 - Stack you are using (including the version): nodejs-express current

I considered a `USER node` is needed before the `RUN npm install` in the `user-app/` directory, but even that would create a node_modules with a uid != to mine. Perhaps the volume can be mounted in a way that uid mapping occurs? I'm not sure, does something similar happen in other stacks?

sam-github:
Maybe I should report this as a seperate bug, but is related. The `npm install` runs as root in the container, and since npm runs package-defined scripts, it has a protective mechanism to avoid running them as root, it drops its privileges to "nobody". This goes quite poorly when the package install has to write artifacts to disk, like every C++ addon does.

Reproduction:
```
appsody init nodejs-express
npm install modern-syslog
appsody run
cloud-node/ex-ext2 % appsody run
Running development environment...
Using local cache for image dev.local/nodejs-express:0.3.0                             
[...]

PR #739, opened 2020-03-27
Language agnostic demo of root-owned dep bug

sam-github:
Issue: https://github.com/appsody/stacks/issues/518

Description: https://github.com/appsody/stacks/issues/518#issuecomment-601957506

This is a minimal, language agnostic, reproduction of issue #518.

Once it works, it might be worth cleaning up and merging as a language
agnostic example of satisfying the non-root user certification
requirement.

To reproduce:
- checkout this PR branch
- cd incubator/starter
- `./test-appsody-test` (fails)
- `./test-appsody-test-root-leakage` (fails)
- `./test-appsody-run` (fails)
- `./test-appsody-build` (passes)

### DIFF
--- a/incubator/starter/image/Dockerfile-stack
+++ b/incubator/starter/image/Dockerfile-stack
@@ -8,7 +8,7 @@ FROM registry.access.redhat.com/ubi7/ubi
 ENV APPSODY_MOUNTS=/:/project/userapp
 
 # Define a volume for dependencies to be stored between runs, for example
-ENV APPSODY_DEPS=/project/deps
+ENV APPSODY_DEPS=/project/userapp/deps
 
 # Directories to watch for changes in
 ENV APPSODY_WATCH_DIR=/project/userapp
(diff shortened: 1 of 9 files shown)


incubator/starter/image/Dockerfile-stack
```diff
@@ -60,3 +60,8 @@ WORKDIR /project
 
 # Expose the relevant ports (change this to be specific to your application).
 EXPOSE 8080
+
+# Run as a non-root user in group 0.
+RUN useradd -u 1001 -r -g 0 -s /usr/sbin/nologin -m default \
```
groeges: It may be better to not specify a uid ie 1001 to the userid you are creating and to let the system assign this, as you do not know what base images people will be using and whether uid 1001 is already being used. You can still do all the operations needed but using the userid ie default, instead of the uid value.

sam-github: This isn't really relevant to the PRs purpose - which is to demonstrate a blocking issue we are having with appsody.

That said, you make a good point. Do you work on the java stacks? This code was essentially copied from https://github.com/appsody/stacks/blob/a3be05c738df79cd99138096c4a21d27778bb50b/incubator/java-spring-boot2/image/Dockerfile-stack#L8 and https://github.com/appsody/stacks/blob/a3be05c738df79cd99138096c4a21d27778bb50b/incubator/java-openliberty/image/Dockerfile-stack#L3  It would help stack authors to have clear guidelines on how to meet the non-root requirement.

PR #746, opened 2020-04-07
Move OL fixpack to 20.0.0.3

### LINKED ISSUE #740: java-openliberty stack not replacing variable inside server.xml with environment variable
**Describe the bug**
I am updating this tutorial to switch from using the java-microprofile application stack to the java-openliberty stack: https://developer.ibm.com/tutorials/connecting-appsody-java-microprofile-applications-to-jdbc-data-sources/

The one thing that keeps giving me trouble, and only inside a container in a pod, is that the resolution of variables with the value of environment variables is not working.

This problem does not happen with `appsody run` or with reading values from bootstrap.properties. It only happens with reading environment variables when the code is running inside a pod container after  `appsody deploy`

Whenever the REST endpoint for accessing the database is hit, the request fails and the following message is issued in the logs:
```
Caused by: java.sql.SQLNonTransientException: [jcc][t4][10380][11951][4.25.13] Required property "${db_server}" is unknown host. ERRORCODE=-4222, SQLSTATE=08001 DSRA0010E: SQL State = 08001, Error Code = -4,222
```
Note that it says "${db_server}" instead of the actual value of the variable.

If I ssh into the container in the pod, I can see the `db_server` environment variable set in the container, and if I change anything in the server.xml file, such as modifying an unrelated value, that change forces the server to restart and then it somehow performs the variable replacement correctly and everything works fine.

I went as far as adding a statement like `System.out.println("DB Server:" + System.getenv("db_server"));` in the source code to prove that the `db_server` environment variable is set before the server starts and that statement prints the correct value right before the exception, so it seems somehow the Liberty server is "seeing" the wrong value even though the JVM has the correct value.

**To Reproduce**
1. Clone the "java-openliberty" branch of this application: https://github.com/IBM/icpa-jee-tutorial/tree/java-openliberty
2. Create a secret in the cluster:
```
cat<<EOF > .db2.temp.env
host=doesntmatter.liberty.fails.before.this
port=50001
user=doesntmatter
password=doesntmatter
db=BLUDB
EOF

kubectl create secret generic sample-jee-db-secret --from-env-file=.db2.temp.env
```

3. Request the response:
```
curl host:port/starter/database
```

```
Caused by: java.sql.SQLNonTransientException: [jcc][t4][10380][11951][4.25.13] Required property
 "${db_server}" is unknown host. ERRORCODE=-4222, SQLSTATE=08001 DSRA0010E: SQL State = 
08001, Error Code = -4,222
```

**Expected behavior**
The ${db_server} variable should be resolved to `doesntmatter.liberty.fails.before.this` and the server should complain about not finding that bogus server name (it is valid in the real case, but that is besides the point)

**Actual behaviour**
Somehow, the ${db_server} variable is not resolved. 

**Environment Details (please complete the following information):**
 - Kubernetes on Docker Desktop and Kubernetes on IBM Cloud

**If applicable please specify:**
 - CLI version: appsody 0.5.9
 - Stack you are using (including the version):
incubator/java-openliberty 0.2.3

**Screenshots**

**Additional context**
If I `ssh` into the container and modify some other constant in the `server.xml` file directly in the container, the Open Liberty re-reads the server.xml file and then the REST endpoint works without issue.

I read about this issue (https://github.com/OpenLiberty/open-liberty/issues/10674) and wondered whether it is possible the server cached the wrong value upon startup.

*Update post initial creation of the issue*:I just found out that If I replace ${db_server} with ${env.db_server}, then my code works! Obviously still a serious regression (relative to java-microprofile) that will cost people a lot of time, but at least I can make progress.

scottkurz:
Signed-off-by: Scott Kurz <skurz@us.ibm.com>

### Checklist:

- [ X] Followed the [commit message guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md#commit-message-guidelines).

- [X ] Stack adheres to [Appsody stack structure](https://github.com/appsody/website/blob/master/content/docs/stacks/stacks-overview.md#stack-structure).

- [X ] Updated the stack version in `stack.yaml`

### Related Issues:
* Fixes #740 
* Updates Open Liberty version to Fix Pack 20.0.0.3


scottkurz:
@Kamran64 @EnriqueL8   - ready to merge.  Thank you.

PR #765, opened 2020-04-17
quarkus: add `kafka` template.

ianpartridge:
Add a new `kafka` template to the Quarkus stack.

This template demonstrates the use of MicroProfile Reactive Messaging to produce and consume messages via Kafka, with Quarkus.  It is inspired by the Quarkus Kafka guide - https://quarkus.io/guides/kafka

Documentation is included showing how to deploy Kafka locally and on Kubernetes.


ianpartridge:
Hi @paulrobinson - please take a look!

paulrobinson:
Thanks @ianpartridge I'll take a look.

My immediate concern, is how we will maintain these code templates. I got bitten by this when I recently updated the Quarkus version. There was a change to the quickstart, I based the default template on, which I didn't notice. It was a breaking change, so I needed to spend a while figuring out why it didn't work anymore.

However, i'm not sure how we can avoid this problem. Maybe we just try to keep the code as similar as possible to the quickstart, such that a diff can be done manually each time the Quarkus version is updated? When i review this PR, I'll do a diff against the quickstart and see how feasible it is to do that.

ianpartridge:
Thanks Paul, that's a fair concern. In the case of this template, the code is deliberately a straight copy-paste to aid in keeping things in sync.

paulrobinson:
Ok great. Should be easy-ish to maintain then. I was able to fix the deafault one pretty quickly once I thought to check the diff.

I suppose the real issue comes if we end up wanting all/most of the guide examples in here. But that's probably a worry for later, if the problem ever arrises.

incubator/quarkus/README.md
```diff
@@ -12,7 +12,44 @@ See: https://quarkus.io/
...
+
+A bean produces a random `Integer` onto a Reactive Messaging channel named "generated-price" every 5 seconds. This channel is connected to a Kafka topic named "prices". Another bean consumes from that topic, applies a conversion, and publishes a `double` onto another channel named "my-data-stream". Finally, that channel is streamed to a JAX-RS endpoint `/prices.html`, using [Server-sent Events](https://en.wikipedia.org/wiki/Server-sent_events).
+
+#### Running locally
+
+The template provides a sample `docker-compose.yaml` which you can use to run a single-node Kafka cluster on your local machine. To start Kafka locally, run `docker-compose up`. This will start Kafka, Zookeeper, and also create a Docker network on your machine, which you can find the name of by running `docker network list`.
+
+To run the application using Appsody, use this command, substituting in the name of your Docker network:
+
+`$ appsody run --network network_name --docker-options "--env KAFKA_BOOTSTRAP_SERVERS=kafka:9092"`
```
Kamran64, line 33: I can reach the /prices endpoint but I get a `404 - Resource Not Found` on the default `localhost:8080` endpoint. Is that expected?

ianpartridge: Yes that's correct.

incubator/quarkus/README.md
```diff
@@ -12,7 +12,44 @@ See: https://quarkus.io/
...
+
+### `kafka` template
+
+This template provides a Java application which demonstrates using MicroProfile Reactive Messaging to consume and produce messages on Kafka topics. The application uses the example from the [Quarkus Kafka guide](https://quarkus.io/guides/kafka).
+
+A bean produces a random `Integer` onto a Reactive Messaging channel named "generated-price" every 5 seconds. This channel is connected to a Kafka topic named "prices". Another bean consumes from that topic, applies a conversion, and publishes a `double` onto another channel named "my-data-stream". Finally, that channel is streamed to a JAX-RS endpoint `/prices.html`, using [Server-sent Events](https://en.wikipedia.org/wiki/Server-sent_events).
+
+#### Running locally
+
+The template provides a sample `docker-compose.yaml` which you can use to run a single-node Kafka cluster on your local machine. To start Kafka locally, run `docker-compose up`. This will start Kafka, Zookeeper, and also create a Docker network on your machine, which you can find the name of by running `docker network list`.
```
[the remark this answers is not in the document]
ianpartridge: It's `<directory-name>_default`. I don't think this necessarily matches the project name so the user will need to use `docker network list` to be sure.

ianpartridge:
Agreed that manual testing will be needed, to ensure that `appsody run` and `appsody deploy` work as expected for this template.  I've done this testing myself, others should verify.

ianpartridge:
Hi @paulrobinson did you have a chance to look at this? Thanks :)

paulrobinson:
Hi @ianpartridge, sorry I've not been able to as something came up yesterday. I'm hoping to look Thurs AM or Friday.

incubator/quarkus/README.md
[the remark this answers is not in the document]
paulrobinson: Can we tell the user this? I was struggling to figure it out until I noticed this comment.

paulrobinson:
I've gone through this now, and I got it working on my laptop (minus the Kubernetes instructions). Although, I'm yet to do a diff against the original quickstart to verify that it's equivalent (I'll try to move onto that now). 

The `README.md` feels a bit muddled to me. If I choose to go down the Kafka route, I need to start in the 'Getting Started' section (which is past the Kafka section) but only follow the first section to initialise the project. I also need to know to specify the Kafka template, which I don't think is mentioned. I think I then follow the instructions in the "`kafka` template" section. But at the end, I'm not sure how to test it, as the instructions are different to testing the default template. I think I go to http://0.0.0.0:8080/prices/stream and observe the stream of prices. 

Would it make sense to separate the README.md into a `defualt` and a `Kafka` version? There would be some duplication, but it would flow better for the user once they have picked the route they want to take. This could scale better if/when we add more templates too (but wouldn't want to prematurely optimise for that now).

ianpartridge:
Thanks for reviewing!  You should be able to go to `prices.html` to see the "UI" :)

I agree about the README.. I'll try and rework it to be a bit clearer but I think we should stick to 1 file for now, until more templates arrive...

paulrobinson:
Keeping in one file does sound ideal. I suppose my main suggestion was to provide a more linear path for the user regardless of which template they choose (at the cost of duplication).

incubator/quarkus/templates/kafka/docker-compose.yaml
```diff
@@ -0,0 +1,30 @@
+version: '2'
+
+services:
+
+  zookeeper:
+    image: strimzi/kafka:0.15.0-kafka-2.3.1
```
paulrobinson, line 6: Was here a reason why you upgraded this version? It seems to work, but it's a change from what's in the original quickstart. See the version targeting Quarkus 1.3.0.Final: https://github.com/quarkusio/quarkus-quickstarts/blob/296582a9ccc2393aa5ec79b2f3e56d87fc7d021d/kafka-quickstart/docker-compose.yaml and the current master version: https://github.com/quarkusio/quarkus-quickstarts/blob/master/kafka-quickstart/docker-compose.yaml. Both reference an older Kafka version.

incubator/quarkus/templates/kafka/docker-compose.yaml
```diff
@@ -0,0 +1,30 @@
...
+      "sh", "-c",
+      "bin/kafka-server-start.sh config/server.properties --override listeners=$${KAFKA_LISTENERS} --override advertised.listeners=$${KAFKA_ADVERTISED_LISTENERS} --override zookeeper.connect=$${KAFKA_ZOOKEEPER_CONNECT}"
+    ]
+    depends_on:
+      - zookeeper
+    ports:
+      - "9092:9092"
+    environment:
+      LOG_DIR: "/tmp/logs"
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:9092
```
paulrobinson, line 28: I see `KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092` in the original QS. Is this hostname change needed to make it work with Appsody?

incubator/quarkus/templates/kafka/mvnw
```diff
@@ -0,0 +1,305 @@
...
+if [ -r "$BASE_DIR/.mvn/wrapper/maven-wrapper.jar" ]; then
+    if [ "$MVNW_VERBOSE" = true ]; then
+      echo "Found .mvn/wrapper/maven-wrapper.jar"
+    fi
+else
+    if [ "$MVNW_VERBOSE" = true ]; then
+      echo "Couldn't find .mvn/wrapper/maven-wrapper.jar, downloading it ..."
+    fi
+    if [ -n "$MVNW_REPOURL" ]; then
+      jarUrl="$MVNW_REPOURL/io/takari/maven-wrapper/0.5.3/maven-wrapper-0.5.3.jar"
```
paulrobinson, line 215: This seems to be an older version than what I see in the QS: https://github.com/quarkusio/quarkus-quickstarts/blob/master/kafka-quickstart/mvnw#L215

Can you point me at where you got the QS code from that you based this template on? I think you need to base it on https://github.com/quarkusio/quarkus-quickstarts/tree/296582a9ccc2393aa5ec79b2f3e56d87fc7d021d/kafka-quickstart which is the 1.3.0.Final tag of the Kafka Quickstart. 

We'll then update the source code in lock-step when the Quarkus version is updated. Note that the Quarkus Quickstarts are released and tested against the equivalent Quarkus version, so we should try to follow that or we could get some unexpected behaviour.

ianpartridge: I think I used the source code from the HEAD of `master` here: https://github.com/quarkusio/quarkus-quickstarts/tree/master/kafka-quickstart

`mvnw` is different because that is copied from the `default` template in this stack. I think we should keep those in sync and update `mvnw` across both templates after this is merged. WDYT?

incubator/quarkus/templates/kafka/docker-compose.yaml
ianpartridge: Yes it's needed so that the `appsody run` container can talk to Kafka over the Docker network that this `docker-compose.yaml` creates.  We've used the same approach in other stacks (Spring, Node.js etc.)

incubator/quarkus/templates/kafka/docker-compose.yaml
ianpartridge: I used the `docker.compose-yaml` that we've used in the Spring and Node.js `kafka` templates. I didn't realise the one in the quickstart was backlevel.  Kafka 2.5 is out now so even this is out of date.

To be honest I don't think it hugely matters - this `docker-compose.yaml` is only going to be used for people who are experimenting with the template. Any real app will likely use a pre-existing Kafka that the developer already has access to.

Kamran64:
@ianpartridge @paulrobinson - Is this critical for today? Can merge if @paulrobinson is happy.

incubator/quarkus/templates/kafka/docker-compose.yaml
paulrobinson: That's true, and it does seem to work for the experimenting.

incubator/quarkus/templates/kafka/docker-compose.yaml
paulrobinson: OK, good to know. Thanks.

ianpartridge:
We'd really like to get this merged today so we can consume it in the RC of another project.

incubator/quarkus/templates/kafka/mvnw
paulrobinson: Good point, the default template won't be exactly in sync now either. 

I agree, it would be good to get `mvnw` updated after the PR goes in. Going forward, I think it would be good to reduce the differences as much as possible, then apply a thorough process of keeping them in sync when the Quarkus version is updated. I might be able to look at doing that later today. When would you need the PR by in order to make the deadline for your release?

Kamran64:
Okay @ianpartridge - Please can you bump the stack version then this can be merged asap.

incubator/quarkus/templates/kafka/mvnw
ianpartridge: Yep let's update `mvnw` afterwards. And also get the stack on Quarkus 1.4 :)

Sorry for the squeeze on getting this in, getting the new template in today would be really helpful for us.

PR #770, opened 2020-04-21
java-openliberty: use ENV variables for security credentials

### LINKED ISSUE #443: Implement service monitoring support for java-openliberty and add timeoutSeconds for readinessProbe
Following https://github.com/appsody/appsody-operator/pull/125, the Appsody operator supports creating service monitor if there is a `spec.monitoring` entry in the `app-deploy.yaml` file. The only requirement to utilise this is a label field, like so:

```
apiVersion: appsody.dev/v1beta1
kind: AppsodyApplication
metadata:
  name: example-appsodyapplication
spec:
  ....
  monitoring:
    endpoints:
    - basicAuth:
        password:
          key: password
          name: pass
        username:
          key: username
          name: user
      interval: 5s
      tlsConfig:
        insecureSkipVerify: true
    labels:
      k8s-app: APPSODY_PROJECT_NAME
```
We introduced this functionality for Python Flask, Node.js-Express, and Node.js-functions but it would be nice to see this for the Open Liberty stack also.

Currently, we also see a Knative error if a stack uses the `readinessProbe` and `timeoutSeconds` is not defined.

**Describe the solution you'd like**
Add a new field (`monitoring`) to the `app-deploy` file underneath `spec`. The java-microprofile stack has authenticated endpoints and therefore requires a `basicAuth` field also.

Also add a new field, `timeoutSeconds` under the `readinessProbe` and assign it an appropriate value e.g. 1.

**Additional context**
Example PR to implement service monitor support into a stack: https://github.com/appsody/stacks/pull/419/files

**Note**: The PR above is for a stack without an authenticated endpoint, see: https://github.com/appsody/appsody-operator/pull/125 for a more suitable approach for microprofile.

### LINKED ISSUE #750: java-openliberty: Use environment variables for username and password in appsody build image
By design, the Open Liberty stack removes the quick-start-security configuration from the build image produced via `appsody build`. This is to prevent accidentally deploying an image with a default user/password which is available for local development.

However, the removal of quick-start-security is not obvious and requires additional configuration to make the deployment useable (i.e. connecting to Prometheus in OpenShift)

**Suggested solution**
We can enhance this by using environment variables for the username/password in the Liberty config. We can configure defaults for these variables in a separate config file and delete THAT rather than the whole quick-start-security config. When the app is deployed, a user can add a secret to the deployment and expose the username/password of that secret as ENV vars to the deployment.

awisniew90:
### Checklist:

- [ x] Read the [Code of Conduct](https://github.com/appsody/website/blob/master/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/appsody/website/blob/master/CONTRIBUTING.md).

1. Use environment variables for quick-start-security username/password
2. Add a "default-credentials.xml" which gets gets removed during `appsody build`
3. Add documentation to show how to enable the HTTPS port and set user credential env vars
4. Add documentation on how to add a Service Monitor to allow Prometheus to authenticate

### Related Issues:
Fixes #750


incubator/java-openliberty/README.md
```diff
@@ -115,6 +116,68 @@ The command `appsody test` launches the Open Liberty server, runs integration te
...
+         key: username
+         name: mySecret
+     - name: STACK_PASSWORD
+     valueFrom:
+       secretKeyRef:
+         key: password
+         name: mySecret
+   ```
+
+   This will set environment variables in the deployment called `STACK_USERNAME` and `STACK_PASSWORD` which map to the username and password of the "mySecret" Secret respectively.
```
scottkurz, line 156: of the "mySecret" Secret 
  =>
of `mySecret`, respectively.

(You mentioned it's a Secret enough I think).

incubator/java-openliberty/templates/default/src/main/liberty/config/server.xml
```diff
@@ -3,6 +3,8 @@
         <feature>microProfile-3.2</feature>
     </featureManager>
 
+    <quickStartSecurity userName="${stack.username}" userPassword="${stack.password}" />
```
scottkurz, line 6: It looks great from the perspective of something new, but we changed the "contract" on configDropins/defaults mid-stream.   

Is this OK?   Can you maybe spell out some more thought about why this shouldn't be a concern?   Is there any way to do this with a file named:  `configDropins/defaults/quick-start-security.xml`?

awisniew90: Not sure what you mean by "contract."  Is the concern that we had a file called quick-start-security.xml and now we dont? Or that we were removing it and now we aren't? Given that this is just a change to the template, not sure if it matters.

scottkurz:
If this is steering us towards a 0.3 update because of back-compatibility with already-init'd 0.2 projects then I think we should slow down and consider doing this via doc, especially since we've talked (maybe just privately though?) about using an Open Liberty service token.    

I'm going to close this PR for now, while we discuss.

PR #783, opened 2020-04-28
quarkus: add smallrye-metrics

ianpartridge:
The Quarkus stack does not include any metrics by default.

This PR adds the `quarkus-smallrye-metrics` dependency to `pom.xml` so that a basic set of metrics are automatically generated and hosted on `/metrics`.

The purpose of this is so that the default app is "Kubernetes ready" - in particular, so that the monitoring features in the Appsody Operator work: https://github.com/application-stacks/runtime-component-operator/blob/master/doc/user-guide.md#Monitoring

### DIFF
--- a/incubator/quarkus/image/project/pom.xml
+++ b/incubator/quarkus/image/project/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>dev.appsody</groupId>
     <artifactId>quarkus</artifactId>
-    <version>0.3.3</version>
+    <version>0.3.4</version>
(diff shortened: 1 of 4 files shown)


ianpartridge:
Hi @paulrobinson - can you take a look please?

paulrobinson:
In general I think we need to be hesitant to bloat the stack out with stuff that most users won't need (Which i assume is why @ianpartridge raised the question). However, in this case I think it's a common enough feature to need in a Cloud Native deployment, which is the primary target for Quarkus.

PR #802, opened 2020-05-04
Update README page for Quarkus

arthurdm:
I noticed that a few places the README page for Quarkus still referenced `experimental`, even though it had been moved to the `incubator`.


arthurdm:
I didn't think that this update needed to upgrade the stack version, but please let me know if that should be incremented.  @neeraj-laad @paulrobinson

PR #805, opened 2020-05-06
java-openliberty: fix stack validate

### LINKED ISSUE #799: java-openliberty stack  - kafka template breaks stack validate
Because the kafka template requires supporting containers (kafka, zookeeper), the `appsody test` will fail by default, unless these containers are started and the test run like:  `appsody test --network  kaf2_default`.   These means that `appsody stack validate`, which performs a set of operations against an app init'd from each template, will also fail for `appsody test` for the kafka-template app.

We need to either: 
* Figure out a way to have the test to pass or get skipped by default
* Add a parameter somehow to skip the tests (which would probably need support from appsody CLI to pass through `appsody stack validate`).

Unfortunately, since the test fails because the app deploy fails before the tests are run (rather than the tests themselves failing in execution), we can't simply define zero tests or skip via the failsafe plugin.

matthewpwilson:
### Checklist:

Fix `appsody stack validate` for java-openliberty kafka template.
`appsody test` was failing because the Liberty application did not start. This was because the kafka host specified in `microprofile-config.properties` was `kafka`, the correct hostname for use with the supplied `docker-compose.yaml`.  However, when running `stack validate`, the Docker compose environment wasn't there and therefore the host could not be resolved. 
To fix this I've set the default host for Kafka in `microprofile-config.properties` to be `localhost`.  This allows the Liberty app to start (even though it still won't connect to a Kafka) and therefore `appsody test` will pass during `stack validate`. 
This also means the default configuration is correct for host-mode develepment (`mvn liberty:dev`) so I've removed the profile overrides from `pom.xml`.
I've updated the README, telling the user to set the environment variable to override the Kafka boostrap server when using the Docker compose environment.

I also spotted some unneeded properties in the config file that were causing warning messages.

### Related Issues:
Fixes #799

### DIFF
--- a/incubator/java-openliberty/templates/kafka/README.md
+++ b/incubator/java-openliberty/templates/kafka/README.md
@@ -25,15 +25,13 @@ Start docker compose with the following command:
 
 If you run `docker network list`, you should see a new network with the name of your project directory and the word `_default` appended. For example, `test-appsody-kafka_default`.
 
-Alternatively, if you want to connect to a Kafka broker elsewhere, edit `src/main/resources/META-INF/microprofile-config.properties` and set the value of the `mp.messaging.connector.liberty-kafka.bootstrap.servers` property to the host and port number of the your broker.
-
 ### 3. Run the Appsody application in the new network
 
-Your Appsody application must be run in the same network as Kafka.
+Your Appsody application must be run in the same network as Kafka. You must also supply the host and port number of the Kakfka broker as an environment variable. 
 
 Run the application using the following command:
 
-```appsody run --network test-appsody-kafka_default```
+```appsody run --network test-appsody-kafka_default --docker-options "-e MP_MESSAGING_CONNECTOR_LIBERTY_KAFKA_BOOTSTRAP_SERVERS=kafka:9092"```
 
 ### 4. Produce a message to a topic
 
--- a/incubator/java-openliberty/templates/kafka/README.md
+++ b/incubator/java-openliberty/templates/kafka/README.md
@@ -33,6 +33,14 @@ Run the application using the following command:
 
 ```appsody run --network test-appsody-kafka_default --docker-options "--env MP_MESSAGING_CONNECTOR_LIBERTY_KAFKA_BOOTSTRAP_SERVERS=kafka:9092"```
 
+Alternatively, edit `src/main/resources/META-INF/microprofile-config.properties` as follows:
+
+```
+mp.messaging.connector.liberty-kafka.bootstrap.servers=kafka:9092
+```
+
+(this value is provided in the sample file, commented out)
+
 ### 4. Produce a message to a topic
 
 Run another container in the same network:
--- a/incubator/java-openliberty/templates/kafka/README.md
+++ b/incubator/java-openliberty/templates/kafka/README.md
@@ -27,11 +27,11 @@ If you run `docker network list`, you should see a new network with the name of 
 
 ### 3. Run the Appsody application in the new network
 
-Your Appsody application must be run in the same network as Kafka. You must also supply the host and port number of the Kakfka broker as an environment variable. 
+Your Appsody application must be run in the same network as Kafka. You must also supply the host and port number of the Kafka broker as an environment variable. 
 
 Run the application using the following command:
 
-```appsody run --network test-appsody-kafka_default --docker-options "-e MP_MESSAGING_CONNECTOR_LIBERTY_KAFKA_BOOTSTRAP_SERVERS=kafka:9092"```
+```appsody run --network test-appsody-kafka_default --docker-options "--env MP_MESSAGING_CONNECTOR_LIBERTY_KAFKA_BOOTSTRAP_SERVERS=kafka:9092"```
 
 ### 4. Produce a message to a topic
 
--- a/incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
+++ b/incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
@@ -1,4 +1,5 @@
 mp.messaging.connector.liberty-kafka.bootstrap.servers=localhost:9093
+## mp.messaging.connector.liberty-kafka.bootstrap.servers=kafka:9092
 
 ## incomingTopic1 topic
 mp.messaging.incoming.incomingTopic1.connector=liberty-kafka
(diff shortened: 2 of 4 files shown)


matthewpwilson:
@scottkurz @ianpartridge please review.

incubator/java-openliberty/templates/kafka/README.md
```diff
@@ -25,15 +25,13 @@ Start docker compose with the following command:
 
 If you run `docker network list`, you should see a new network with the name of your project directory and the word `_default` appended. For example, `test-appsody-kafka_default`.
 
-Alternatively, if you want to connect to a Kafka broker elsewhere, edit `src/main/resources/META-INF/microprofile-config.properties` and set the value of the `mp.messaging.connector.liberty-kafka.bootstrap.servers` property to the host and port number of the your broker.
-
 ### 3. Run the Appsody application in the new network
 
-Your Appsody application must be run in the same network as Kafka.
+Your Appsody application must be run in the same network as Kafka. You must also supply the host and port number of the Kakfka broker as an environment variable.
```
ianpartridge, line 30: typo: Kafka

incubator/java-openliberty/templates/kafka/README.md
```diff
@@ -25,15 +25,13 @@ Start docker compose with the following command:
...
-
 ### 3. Run the Appsody application in the new network
 
-Your Appsody application must be run in the same network as Kafka.
+Your Appsody application must be run in the same network as Kafka. You must also supply the host and port number of the Kakfka broker as an environment variable. 
 
 Run the application using the following command:
 
-```appsody run --network test-appsody-kafka_default```
+```appsody run --network test-appsody-kafka_default --docker-options "-e MP_MESSAGING_CONNECTOR_LIBERTY_KAFKA_BOOTSTRAP_SERVERS=kafka:9092"```
```
ianpartridge, line 34: We used `--env` instead of `-e` in the other templates' README, but no biggie.

incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -1,10 +1,8 @@
-mp.messaging.connector.liberty-kafka.bootstrap.servers=kafka:9092
+mp.messaging.connector.liberty-kafka.bootstrap.servers=localhost:9093
```
ianpartridge, line 1: Could you explain why the port changes too?

incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
```diff
@@ -1,10 +1,8 @@
...
 
 ## incomingTopic1 topic
 mp.messaging.incoming.incomingTopic1.connector=liberty-kafka
 mp.messaging.incoming.incomingTopic1.topic=incomingTopic1
-mp.messaging.incoming.incomingTopic1.value.serializer=org.apache.kafka.common.serialization.StringSerializer
```
ianpartridge, line 6: This change seems unrelated?

incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
matthewpwilson: The Kafka has two separate listeners configured for inside the Docker network and outside, so they must listen on two separate ports.  Two listeners are required because we need to advertise different hostnames. Not sure if we need to explain this in the README?

incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
matthewpwilson: I noticed these properties were unnecessary and were causing warnings - you don't need a serializer for an incoming topic and the deserializer will happily default.

incubator/java-openliberty/templates/kafka/README.md
matthewpwilson: Thanks, fixed.

incubator/java-openliberty/templates/kafka/README.md
matthewpwilson: Might as well be consistent

incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
ianpartridge: OK, probably they should be in a separate PR as not related to fixing `appsody stack validate`.

incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
ianpartridge: Ahh that makes sense. I think previously I'd seen Kafka using 9092 for plaintext and 9093 for TLS...

incubator/java-openliberty/templates/kafka/README.md
```diff
@@ -25,15 +25,13 @@ Start docker compose with the following command:
...
-
 ### 3. Run the Appsody application in the new network
 
-Your Appsody application must be run in the same network as Kafka.
+Your Appsody application must be run in the same network as Kafka. You must also supply the host and port number of the Kafka broker as an environment variable. 
 
 Run the application using the following command:
 
-```appsody run --network test-appsody-kafka_default```
+```appsody run --network test-appsody-kafka_default --docker-options "--env MP_MESSAGING_CONNECTOR_LIBERTY_KAFKA_BOOTSTRAP_SERVERS=kafka:9092"```
```
scottkurz, line 34: Maybe instead of this complicated, quoted invocation we could just have it in the **microprofile-config.properties** file commented-out, and instruct them here to edit the file?

incubator/java-openliberty/templates/kafka/src/main/resources/META-INF/microprofile-config.properties
scottkurz: Given the somewhat heavyweight release process,  it's okay with me to combine changes like this as long as you don't mind

ianpartridge: OK OK :) Just a point of principle and git hygiene.

incubator/java-openliberty/templates/kafka/README.md
matthewpwilson: When deploying to a Kube environment I'd expected folks to use the env var, as the broker address is likely to differ between environments. Therefore IMO it makes sense for folks to use the same approach at dev time.  We could offer both options though... I'll add it to the config file and mention in the README

PR #811, opened 2020-05-12
java-openliberty: remove version restriction

awisniew90:
Remove OL version restriction from readme


awisniew90:
@neeraj-laad  - This is ready to merge when you get a chance. I spoke with @Kamran64, and since this is just a minor change to our readme, we dont need to rebuild a new release.

PR #816, opened 2020-05-27
java-openliberty: Work around docker bug

matthewpwilson:
### Checklist:

When trying to do an `appsody build` using the java-openliberty stack in a GitHub Actions, CI environment, I hit the following error: 
```
[Docker] Step 18/68 : COPY --chown=1001:0 --from=compile /configlibdir/ /config
[Docker] failed to export image: failed to create image: failed to get layer sha256:27f55a0113aeec813a6bb31695fd2449c4dc13fe53adb7a67e144c325b63bdf7: layer does not exist
[Error] exit status 1
```
This appears to be caused by moby/moby#37965.  The characteristics of the `COPY` don't exactly match the description in the issue but the symptom is the same.    This issue has been open for a while and there's no sign of a fix so I think it's worth putting a workaround in the stack. I've added the suggested `RUN true`, which fixes the issue for me.  It seems like subsequent `COPY` commands could hit the same issue, but in practice they don't for me, so I've just put in the one instance for now.

### DIFF
--- a/incubator/java-openliberty/image/project/Dockerfile
+++ b/incubator/java-openliberty/image/project/Dockerfile
@@ -64,6 +64,9 @@ FROM openliberty/open-liberty:{{.stack.libertyversion}}-kernel-java8-openj9-ubi
 #2a) copy any resources 
 COPY --chown=1001:0 --from=compile /shared /opt/ol/wlp/usr/shared/
 
+# Work around https://github.com/moby/moby/issues/37965
+RUN true
+
 # 2b) next copy shared library
 #      but can't assume config/lib exists - copy from previous stage to a tmp holding place and test
 COPY --chown=1001:0 --from=compile /configlibdir/ /config
--- a/incubator/java-openliberty/image/project/Dockerfile
+++ b/incubator/java-openliberty/image/project/Dockerfile
@@ -64,7 +64,7 @@ FROM openliberty/open-liberty:{{.stack.libertyversion}}-kernel-java8-openj9-ubi
 #2a) copy any resources 
 COPY --chown=1001:0 --from=compile /shared /opt/ol/wlp/usr/shared/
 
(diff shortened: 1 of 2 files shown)


matthewpwilson:
@scottkurz please review

matthewpwilson:
Thinking about this further I've realised my testing isn't sufficient to prove my fix does the job.... I'll close the PR for now until I've proved it

matthewpwilson:
I believe this is fine - I tested by pushing my new stack image to Dcokerhub, and changed my project's .appsody-config.yaml to point at it.  If I understand correctly Appsody will pick up the new Dockerfile from the stack image and use that for `appsody build`. This works correctly with my fix in place.

incubator/java-openliberty/image/project/Dockerfile
```diff
@@ -64,6 +64,9 @@ FROM openliberty/open-liberty:{{.stack.libertyversion}}-kernel-java8-openj9-ubi
 #2a) copy any resources 
 COPY --chown=1001:0 --from=compile /shared /opt/ol/wlp/usr/shared/
 
+# Work around https://github.com/moby/moby/issues/37965
```
scottkurz, line 67:
```suggestion
# Work around https://github.com/moby/moby/issues/37965.  It seems the problem may be triggered when the previous COPY command adds no contents to the next layer (because it was empty). 
```

scottkurz: @matthewpwilson seems there's some more specific guidance to share here.... which is the fact this is triggered because the previous COPY added nothing.  Not sure I like my wording.. what do you think?

scottkurz:
@Kamran64 we can build this:

> Fix docker bug hit during `appsody build`:  failed to export image: failed to create image: failed to get layer

Thanks

Issue #126, opened 2019-07-30
Errors when importing Node.js Express project into Eclipse


eharris369:
**Describe the bug**
There are errors for the launch.json and tasks.json files when importing a Node.js Express project into Eclipse.  It does not like the comments in both files and the trailing ',' in tasks.json:

- Expected name at 2:4	launch.json	/appsodyNode1/.vscode	line 2	JSON Problem
- Expected name at 2:4	tasks.json	/appsodyNode1/.vscode	line 2	JSON Problem
- Expected value at 48:4	tasks.json	/appsodyNode1/.vscode	line 48	JSON Problem

**To Reproduce**
1. Create an appsody Node.js Express project
2. Using Eclipse IDE for Enterprise Java Developers (https://www.eclipse.org/downloads/packages/release/2019-06/r/eclipse-ide-enterprise-java-developers), import the project into Eclipse using the **File** > **Import** > **General** > **Projects from Folder or Archive** wizard

**Actual behaviour**
Errors in the Markers view for the imported project

**Environment Details (please complete the following information):**
 - OS: Windows 10
 - Browser: N/A
 - Version: latest

**If applicable please specify:**
 - CLI version:
 - Stack you are using:

**Screenshots**

**Additional context**

neeraj-laad:
Thanks @eharris369 for raising this issue.
We will investigate and update the status here shortly.

PR #825, opened 2020-06-03
rust-tide: Move to gdb remote and add vscode support

No9:
### Checklist:

This patch adds the gdb server to the Dockerfile-stack to allow remote debuggers to be attached to the running debug instance. It also adds support for VSCode by adding an additional task to copy the symbols from the container to the local dev environment and provides a launcher for the debugger.
Usage is also documented in the README.


experimental/rust-tide/README.md
```diff
@@ -49,32 +49,28 @@ Templates are used to create your local project and start your development. When
...
+   This is done by pressing `CTRL + SHFT + B` and selecting `Appsody: debug` 
+   Or through the menu option `Terminal --> Run Task --> Appsody: debug`
+   The command will start the the app with the gdb server and wait for incoming connections from any address to port 1234. 
 
-Once in lldb, you can use `breakpoint set` to set breakpoints in your application, and then `run` to start the app.
+2. Run the Appsody copy task to get the symbols for the the local debugger.
+   This is done by pressing `CTRL + SHFT + B` and selecting `Appsody: copy` 
+   Or through the menu option `Terminal --> Run Task --> Appsody: copy`
+    
+3. Now we can debug the program with pressing `F5`
```
Kamran64, line 70: I get this error at this point:
```
Unable to start debugging. Launch options string provided by the project system is invalid. Unable to determine path to debugger.  Please specify the "MIDebuggerPath" option.
```
Have you seen this before?

experimental/rust-tide/templates/default/.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
Kamran64, line 2: This might cause issues when importing projects into other IDE's (https://github.com/appsody/stacks/issues/126). Please could you remove the comments just in case?

experimental/rust-tide/README.md
No9: That looks like you don't have gdb installed I have put notes in the README to point towards how to do this for Linux, Mac and Windows.

experimental/rust-tide/templates/default/.vscode/launch.json
No9: Removed with the new PR.

experimental/rust-tide/README.md
EnriqueL8: Have you tried using `CodeLLDB`? We had some issues with it where the process was still running in the container after the debug session in Vscode was finished. Did you encounter the same issue, and is this the solution to get around it? If so it might be good to update the other Rust stack to be consistent

No9: Hey @EnriqueL8 
Yes I recreated the issue with `CodeLLDB` 
I also looked at using lldb connected directly to ensure the VSCode wasn't causing an issue. 
Despite the simpler approach the issues continued along with a general inability to step through the process execution and some random crashes.
I don't think LLDB remote debugging is up to it at the moment hence taking the gdbserver approach.
Also note that the gdbserver approach is not perfect but it is a lot more robust and usable than lldb remote.
FYI I also evaluated installing https://www.gdbgui.com/ into the container.
This again this wasn't perfect but it did remove the need to integrate with external developer tools but meant the developer would need to use a none familiar tool.
I think the implementation in this PR gives the most flexibility but I'll always open to reconsidering it if users have different preferences.

EnriqueL8: Nice, thanks for the info! Might be good documenting this somewhere

PR #832, opened 2020-06-10
nodejs-express: Shut down Kafka clients when HTTP server is closed

### LINKED ISSUE #831: nodejs-express: kafka template tests hang if kafka is reachable
**Note:** I'm raising this retrospectively having investigated and found a solution.  I'll raise a PR to illustrate how to resolve this issue.

**Describe the bug**
The `kafka` template for the `nodejs-express` stack creates a producer and consumer and tries to connect to a Kafka server based on the `KAFKA_BOOTSTRAP_SERVERS` env var.

If a Kafka is reachable, then the producer and consumer will connect when the app is initialized during the stack tests.  When the stack tests complete, they will call `close` on the HTTP server, however because there are still Kafka clients connected, the test process will never terminate.

**To Reproduce**
You can do a straightforward recreate as follows:
```
mkdir testkafka && cd testkafka && appsody init nodejs-express kafka
appsody test
```
As there is no local kafka running, this will eventually terminate as the kafka broker connections time out.
If you stand up a local kafka and then run, it'll hang:
```
docker-compose up -d
appsody test --network testkafka_default --docker-options "--env KAFKA_BOOTSTRAP_SERVERS=kafka:9092"
```
**Expected behavior**
Tests should pass, and then the `appsody test` command should complete.

**Actual behaviour**
Tests pass, but then `appsody test` hangs (never completes).

djones6:
This PR illustrates how to avoid a hang when running the stack tests where the application makes connections to external services.

### Checklist:

I have added a callback on the HTTP server `close` event, to disconnect the Kafka producer and consumer.  This allows the process to terminate normally.

### Related Issues:

Resolves #831

### DIFF
--- a/incubator/nodejs-express/templates/kafka/app.js
+++ b/incubator/nodejs-express/templates/kafka/app.js
@@ -31,8 +31,15 @@ module.exports = (options) => {
     // 'log.connection.close' : false,
   }
 
-  Produce(Kafka, config, log, app);
-  Consume(Kafka, brokers, log);
+  producer = Produce(Kafka, config, log, app);
+  consumer = Consume(Kafka, brokers, log);
+
+  // Disconnect our Kafka clients when the HTTP server is stopped.
+  options.server.on('close', function () {
+    log.info('Disconnecting Kafka clients');
+    producer.disconnect();
+    consumer.disconnect();
+  });
 
   return app;
 };
@@ -89,6 +96,7 @@ function Produce(Kafka, config, log, app) {
 
   // starting the producer
   producer.connect();
+  return producer;
 }
 
 // Based on:
@@ -145,4 +153,5 @@ function Consume(Kafka, brokers, log) {
 
   // starting the consumer
   consumer.connect();
+  return consumer;
 }
--- a/incubator/nodejs-express/stack.yaml
+++ b/incubator/nodejs-express/stack.yaml
@@ -1,5 +1,5 @@
 name: Node.js Express
-version: 0.4.10
+version: 0.4.11
 description: Express web framework for Node.js
 license: Apache-2.0
 language: nodejs
--- a/incubator/nodejs-express/stack.yaml
+++ b/incubator/nodejs-express/stack.yaml
@@ -1,5 +1,5 @@
 name: Node.js Express
-version: 0.4.11
+version: 0.4.12
 description: Express web framework for Node.js
 license: Apache-2.0
 language: nodejs


incubator/nodejs-express/templates/kafka/app.js
```diff
@@ -31,8 +31,15 @@ module.exports = (options) => {
...
 
-  Produce(Kafka, config, log, app);
-  Consume(Kafka, brokers, log);
+  producer = Produce(Kafka, config, log, app);
+  consumer = Consume(Kafka, brokers, log);
+
+  // Disconnect our Kafka clients when the HTTP server is stopped.
+  options.server.on('close', function () {
+    log.info('Disconnecting Kafka clients');
+    producer.disconnect();
```
gireeshpunathil, line 40:
```suggestion
    if (producer._isConnected)
      producer.disconnect();
```

just in case if the close call comes too fast before the producer is setup? I am not sure if this is possible in real world cases, but I got `Producer not connected` error while trying to isolate the issue, with a similar change

djones6: I don't like the look of checking an internal variable called `_isConnected` - this looks like a high risk of breakage in the future?

As I understand it, if the producer / consumer are currently connected, then a call to `disconnect()` will disconnect them.  If they failed to connect for some reason, this call would be unnecessary (but do no harm since we're shutting down anyway).

The situation that might be a problem would be if the clients were still in the process of connecting, and were going to connect successfully, but we called disconnect before they finished.  I don't know how to test that, and I didn't find any documentation on how you might handle that.  I can't imagine the behaviour would be any worse than the current (indefinitely hanging) behaviour though.

gireeshpunathil: I don't like the use of internal variable either. I saw a crash, and I figured the root cause is a race condition, and hence suggested. I am not blocking the PR on this.

djones6:
@gireeshpunathil Do you think this can be merged?

gireeshpunathil:
> @gireeshpunathil Do you think this can be merged?

Yes. @neeraj-laad - for your review 

(Ironically, I am a maintainer of the stack, but not a maintainer of this repo)

Kamran64:
@gireeshpunathil - Usually we wait to see an approval from the stack maintainer before looking.
@djones6 - Please could you bump the stack version in the `stack.yaml` and then we can get this merged. Thanks!

gireeshpunathil:
> @gireeshpunathil - Usually we wait to see an approval from the stack maintainer before looking.

can you pls document this? It is neither evident here, nor an established practice in OSS

neeraj-laad:
@gireeshpunathil I've added you as a stack maintainers group. 

Also the review for maintainers is documents in our contribution process.

neeraj-laad:
@djones6 please can you bump the stack version so we can merge this?

PR #834, opened 2020-06-12
python-flask: Avoid permission errors when building dependencies

### LINKED ISSUE #833: python-flask: app dependencies not being included due to clash of UIDs
**Describe the bug**
The python stack (0.2.3) already ensures the application does not run as root (the Docker creates a user called "worker" for this). A recent change also set the env variable APPSODY_USER_RUN_AS_LOCAL, which causes the docker file to be run as the local user...but this clashes with the UID of worker. 

**To Reproduce**
1. Create a new project, initing with the python-stack
2. appsody run
3. ...you will notice a line that says "/bin/sh: 1: pipenv: Permission denied" in the output, which means that piping failed to run and pull in any new dependencies defined by the local app. This failure is due to the mismatched UIDs

henrynash:
Avoid clashing UIDs in the attempt to ensure the application does not run as root. Since the Dockerfiles already create a non-root user, do not also set APPSODY_USER_RUN_AS_LOCAL to true.

Attempting to only use APPSODY_USER_RUN_AS_LOCAL=true and not create a user in the Dockerfile, leads to different issues regarding clashing UIDs

Fixes #833

### DIFF
--- a/incubator/python-flask/image/Dockerfile-stack
+++ b/incubator/python-flask/image/Dockerfile-stack
@@ -20,7 +20,7 @@ ENV APPSODY_USER_RUN_AS_LOCAL=false
 ENV APPSODY_WATCH_DIR=/project/userapp
 ENV APPSODY_WATCH_REGEX="^.*.py$"
 
-ENV APPSODY_PREP="cd /project/userapp;pipenv lock -r > requirements.txt;python -m pip install -r requirements.txt -t /project/deps"
+ENV APPSODY_PREP="cd /project/userapp;whoami;ls -al;pipenv lock -r > requirements.txt;python -m pip install -r requirements.txt -t /project/deps"
 
 ENV APPSODY_RUN="python -m flask run --host=0.0.0.0 --port=8080"
 ENV APPSODY_RUN_ON_CHANGE=$APPSODY_RUN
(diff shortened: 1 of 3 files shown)


henrynash:
So this is super puzzling. The stack validate fails when running our ci tests (with a permissions error when the docker file is running)....but if I run stack validate locally it works fine. I think this was the original reason by an attempt was done to add RUN_AS_LOCAL. But for the life of me, I cannot work our what is different in our ci environment vs running locally.

henrynash:
So the issue is that when we run under Travis, the volume out of the user app directory shows up in the container as UID 2000 (the Travis UID), which is different to the local user defined in the Dockerfile...and so when the local user tries to run the PREP command, it doesn't have permission to write file into the mounted volume in the container. Mulling a fix....

henrynash:
So the underlying reason appears to be a difference between how docker bind mounting works on macOS (at least when using docker desktop for Mac) and Linux . On macOS, the new osxfs work in Docker desktop appears to change how the permissions of volume mounts are handled. In particular and volume bind mount appears to honour the user currently run in the container (e.g. set by the USER commands). This is country to what the Docker standard says, where a volume bind mount has the UID of the original source point (usually the local file system). This is why under travis, the mounted dir is owned by the travis user (UID 2000), while testing locally the mounters dir was owned by the user created in the Docker container (worker, UID 1000).

I'll work up a fix, that should work on all platforms.

henrynash:
I have tested the proposed fix on macOS and Linux, as well as confirmed that this does work against https://github.ibm.com/vaccine-cold-chain/reeferiotsimulator, which was where this problem was found

PR #841, opened 2020-07-07
nodejs: Add 'build' and 'prune' lifecycle events

djones6:
### Checklist:

I added `npm build` and `npm prune` steps to the build process for the production image, to provide a point of customization in cases where the project needs to perform build actions (such as running the Angular or Typescript compiler).

The 'prune' step is required to remove any development dependencies, and/or perform any cleanup of artifacts generated during the build, which would otherwise bloat the final image. 

Although there is an `npm prune` command, it does not behave the same way as `install`, `test` etc in that it does not invoke any scripts in package.json (none of `prune`, `preprune`, `postprune` are run).  The only way to run a script called `prune` is to run `npm run prune` (which is what I've gone for).  It is run with `--production` so that the user's script can invoke `npm prune` (possibly in addition to other actions), and this will have the effect of running `npm prune --production`, which removes any development-only dependencies.

Update: I added libssl1.1 and ca-certificates.  This is required in order for the service to connect to others over https (with certificate verification), which I feel is a common enough requirement to be part of the base stack.

### DIFF
--- a/incubator/nodejs/README.md
+++ b/incubator/nodejs/README.md
@@ -42,7 +42,7 @@ Templates are used to create your local project and start your development. When
 
 ## Customizing the build
 
-Simple Node projects do not require a 'build' process: they are executed directly from source.  However, some technologies such as Typescipt require commands to be invoked to prepare the project ready to be run.
+Simple Node projects do not require a 'build' process: they are executed directly from source.  However, some technologies such as TypeScript require commands to be invoked to prepare the project ready to be run.
 
 If your project's dependencies require an additional build step, you can do so as follows:
 
--- a/incubator/nodejs/README.md
+++ b/incubator/nodejs/README.md
@@ -40,6 +40,20 @@ Templates are used to create your local project and start your development. When
 
     **NOTE:** Currently the `appsody deploy` command only works for deploying web applications.
 
+## Customizing the build
+
+Simple Node projects do not require a 'build' process: they are executed directly from source.  However, some technologies such as Typescipt require commands to be invoked to prepare the project ready to be run.
+
+If your project's dependencies require an additional build step, you can do so as follows:
+
+- In your `package.json`, edit the `"scripts"` section and define two new scripts:
+  - `"build": "npm install && <build commands>"`
+  - `"prune": "<cleanup commands> && npm prune"`
+
+When initially launching your app using `appsody run`, or building the production image with `appsody build`, the 'build' script is executed (if one exists).  Your build script should invoke `npm install` if any dependencies specified in your `devDependencies` are required as part of the build process - such as command line tools.
+
+When generating the production image, the 'prune' script is executed after that, if it exists, and is run with the `--production` flag.  This allows the removal of development dependencies that were required during the 'build' step.  This could be as simple as running the `npm prune` command, which inherits the production flag, and restores the dependencies to a production state.  You may also use this step to clean up any build artifacts that are not necessary in your production image.
+
 ## License
 
 This stack is licensed under the [Apache 2.0](./image/LICENSE) license
--- a/incubator/nodejs/image/project/Dockerfile
+++ b/incubator/nodejs/image/project/Dockerfile
@@ -8,7 +8,19 @@ RUN rm -rf /project/user-app/node_modules && mkdir -p /project/user-app/node_mod
 
 # Install user-app dependencies
 WORKDIR /project/user-app
-RUN npm install --production 
+RUN npm install --production
+
+# Run a build phase. Projects that need to execute build commands can customize the
+# 'build' script in their package.json. The build script should call 'npm install'.
+# If no build is required, just install production dependencies.
+RUN npm run build --if-present
+
+# Uninstall dev dependencies, leaving only production dependencies. Projects can
+# customize the 'prune' script in their package.json.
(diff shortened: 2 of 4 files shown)


incubator/nodejs/image/project/Dockerfile
```diff
@@ -8,14 +8,29 @@ RUN rm -rf /project/user-app/node_modules && mkdir -p /project/user-app/node_mod
 
 # Install user-app dependencies
 WORKDIR /project/user-app
-RUN npm install --production 
+RUN npm install --production
+
+# Run a build phase. Projects that need to execute build commands can customize the
+# 'build' script in their package.json. The build script should call 'npm install'.
+# If no build is required, just install production dependencies.
```
neeraj-laad, line 15: > The build script should call 'npm install'.
> If no build is required, just install production dependencies.

Why is this needed is we are running `npm install --production` ourselves on line 11?

incubator/nodejs/image/project/Dockerfile
```diff
@@ -8,14 +8,29 @@ RUN rm -rf /project/user-app/node_modules && mkdir -p /project/user-app/node_mod
...
+RUN npm run build --if-present
+
+# Uninstall dev dependencies, leaving only production dependencies. Projects can
+# customize the 'prune' script in their package.json.
+# Ideally this would just be 'npm prune', but this command does not have pre/post
+# hooks. Instead, the user's 'prune' script can itself call 'npm prune' in addition
+# to any additional actions required.
+RUN npm run prune --production --if-present
```
neeraj-laad, line 23: Again are we delegating specific tasks to developer now? will it work fine if I bring my own app that does not have these scripts defined?

incubator/nodejs/image/project/Dockerfile
```diff
@@ -8,14 +8,29 @@ RUN rm -rf /project/user-app/node_modules && mkdir -p /project/user-app/node_mod
...
 
 # Creating a tar to work around poor copy performance when using buildah 1.9.0
 RUN cd / && tar czf project.tgz project
 
 # Copy the dependencies into a slim Node docker image
 FROM node:12-slim
 
+# Install common dependencies (TLS and CA support)
+RUN apt-get update && apt-get install -y libssl1.1 ca-certificates && apt-get clean
```
neeraj-laad, line 32: In the past we've pulled out any apt get updates/upgrades with a view that we want builds to be reproducible. 

Could we install specific versions of these packages?

incubator/nodejs/image/project/Dockerfile
djones6: The idea behind these changes was to leave the build for existing apps unchanged - projects that do not define a 'build' and 'prune' script would build as before.

If a build script is created, and the build process requires additional dependencies (defined in `devDependencies`), they would need to be installed at this stage.  The existing `npm install --production` would not install the `devDependencies` that may be necessary for the build phase (for example, a Typescript project would only require the tsc compiler at build time, which emits standard Javascript for runtime).

incubator/nodejs/image/project/Dockerfile
djones6: There is no change required for existing apps - the `--if-present` flag allows this command to succeed (do nothing).

If you bring an app that worked in the existing nodejs stack, it will continue to work in this version.  If you bring an app that requires a build step (such as running the `tsc` compiler), you would now be able to use it with this stack by doing something like this in your `package.json`:
```
    "build": "npm install && tsc",
    "prune": "npm prune"
```
This works because `npm run build` is run without the `--production` flag, and so calling `npm install` will install the `devDependencies` defined in the package.
When `npm run prune` is run, the `--production` flag sets environment which propagates to the `npm prune` command.  This removes any dependencies that would not be installed by `npm install --production`.

incubator/nodejs/image/project/Dockerfile
ianpartridge: `FROM node:12-slim` is not reproducible anyway though, is it? That image will be updated regularly by upstream.

djones6: Similar to what Ian just commented, I'm not sure it would make sense to install more specific versions than these.  This would match (for example) the packages that would be published in the `node:12` (full fat) image.  We'd want to pick up security fixes to all packages automatically, and as I understand it, the way to do this is to regularly republish the image.

If there's a desire to consume those updates in a more selective / scheduled way, I think that's a separate issue and probably needs some tooling around it (like an intermediate slim image that a stack is based on, with its own release / versioning process).

(as an aside, we're also installing these packages in the nodejs-express stack: https://github.com/appsody/stacks/blob/master/incubator/nodejs-express/image/project/Dockerfile#L29 )

incubator/nodejs/README.md
```diff
@@ -40,6 +40,20 @@ Templates are used to create your local project and start your development. When
 
     **NOTE:** Currently the `appsody deploy` command only works for deploying web applications.
 
+## Customizing the build
+
+Simple Node projects do not require a 'build' process: they are executed directly from source.  However, some technologies such as Typescipt require commands to be invoked to prepare the project ready to be run.
```
BethGriggs, line 45: nit: Typescrip -> TypeScript

